PR #152, opened 2021-06-10
Feature multitenant

osallou:


Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -337,6 +367,9 @@ public function pendingusereditAction($id_space, $id) {
             $modelSpace->setUserIfNotExist($pendingInfo["id_user"], $id_space, $form->getParameter("role"));
             $modelPending->validate($id, $_SESSION["id_user"]);
 
+            // multi-tenant feature: deleting user from core_pending_accounts table for this space in db
+            $modelPending->deleteByPendingAccountId($id_space, $id);
```
osallou, line 371: invalid number of params
should not delete anyway, but use validated (or not) field afterward

docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+version: "3"
```
osallou, line 1: Il ne faut pas commiter le docker compose

Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -48,7 +48,7 @@ public function indexAction($id_space, $letter = "A", $active = "") {
                 $letter = $_SESSION["user_last_letter"];
             }
             else{
-                $letter = "A";
+                $letter = "All";
```
osallou, line 51: why settings to all by default?

Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -337,6 +352,9 @@ public function pendingusereditAction($id_space, $id) {
             $modelSpace->setUserIfNotExist($pendingInfo["id_user"], $id_space, $form->getParameter("role"));
             $modelPending->validate($id, $_SESSION["id_user"]);
 
+            $mailParams = ["id_space" => $id_space, "id_user" => $pendingInfo["id_user"]];
+            $this->notifyUserByEmail($mailParams, "accept");
```
osallou, line 356: accept is a bit too generic name

Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -349,4 +367,57 @@ public function pendingusereditAction($id_space, $id) {
...
+     */
+    public function notifyUserByEmail($params, $origin) {
+        $lang = $this->getLanguage();
+        $fromName = "Platform-Manager";
+        $mail_from = getenv('MAIL_FROM');
+        $from = (!empty($mail_from)) ? $mail_from : "support@platform-manager.com";
+        if ($origin === "add account") {
+            $fromName = "Platform-Manager";
+            $toAdress = $params["email"];
+            $subject = CoreTranslator::Account($lang);
```
osallou, line 405: in general, mails sent from a space should indicated space info , kinda:

Subject: [pfm: Myspacename] bla bla

to differentiate emails from different spaces

Modules/core/Controller/CoretilesController.php
```diff
@@ -82,13 +85,20 @@ public function showMainSubMenu($id){
         
         $lang = $this->getLanguage();
         $modelCoreConfig = new CoreConfig();
+
+        $userSpaces = $this->getUserSpaces();
+        Configuration::getLogger()->debug("IN CONTROLLER", ["adminOf" => $userSpaces['spacesUserIsAdminOf']]);
```
osallou, line 90: bad comment in log.... not explicit

Modules/core/Controller/CoretilesController.php
```diff
@@ -99,4 +109,91 @@ public function docAction(){
...
+
+    /**
+     * 
+     * Manage actions resulting from user request to join or leave a space
+     *
+     * @param int $id_space
+     * @param bool $isMemberOfSpace
+     */
+    public function joinSpaceAction($id_space, $isMemberOfSpace) {
+        Configuration::getLogger()->debug("in DELETING", ["isMemberOfSpace: " => $isMemberOfSpace, "id_space: " => $id_space]);
```
osallou, line 162: bad comment descrption

Modules/core/Controller/CoretilesController.php
```diff
@@ -99,4 +109,91 @@ public function docAction(){
...
+     *
+     * @param int $id_space 
+     */
+    public function NotifyAdminForJoinRequest($id_space) {
+        $lang = $this->getLanguage();
+        $spaceModel = new CoreSpace();
+        $emailSpaceManagers = $spaceModel->getEmailsSpaceManagers($id_space);
+        
+        $mailer = new MailerSend();
+        $mail_from = getenv('MAIL_FROM');
```
osallou, line 188: same comment as before for env var, use config

Modules/core/Controller/CoretilesController.php
```diff
@@ -99,4 +109,91 @@ public function docAction(){
...
+        $this->redirect("coretiles");
+    }
+
+    /**
+     * 
+     * Send an Email to space managers (status > 2) notifying that logged user requested to join space
+     *
+     * @param int $id_space 
+     */
+    public function NotifyAdminForJoinRequest($id_space) {
```
osallou, line 182: should you not refactor, like for user, a method to send different messages?

Modules/core/Model/CorePendingAccount.php
```diff
@@ -19,20 +25,45 @@ public function validate($id, $validated_by){
...
+        $sql = "UPDATE core_pending_accounts SET validated=?, date=?, validated_by=? WHERE id=?";
+        $this->runRequest($sql, array(0, date('Y-m-d'), $validated_by, $id));
+    }
     
     public function add($id_user, $id_space){
         $sql = "INSERT INTO core_pending_accounts (id_user, id_space, validated) VALUES (?,?,?)";
         $this->runRequest($sql, array($id_user, $id_space, 0));
     }
-    
+
```
osallou, line 38: empty spaces

Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -91,14 +91,17 @@ public function indexAction($id_space, $letter = "A", $active = "") {
...
+            $userSpaceInfo = $modelSpaceUser->getUserSpaceInfo2($id_space, $usersArray[$i]["id"]);
             $usersArray[$i]["date_convention"] = CoreTranslator::dateFromEn($userSpaceInfo["date_convention"], $lang);
             $usersArray[$i]["convention_url"] = $userSpaceInfo["convention_url"];
             $usersArray[$i]["date_contract_end"] = CoreTranslator::dateFromEn($userSpaceInfo["date_contract_end"], $lang);
 
             $usersArray[$i]['spaces'] = $modelSpace->getUserSpacesRolesSummary($usersArray[$i]['id']);
         }
 
+        // multi-tenant feature: get only users of this space
+        $usersArray = $this->getUsersOfSpace($id_space, $usersArray);
```
osallou, line 103: missing optimization

Modules/core/Controller/CoretilesController.php
```diff
@@ -99,4 +109,91 @@ public function docAction(){
...
+        $userPendingSpaceIds = array();
+
+        if ($data && count($data) > 0) {
+            foreach ($data as $space) {
+                array_push($userPendingSpaceIds, $space["id_space"]);
+            }
+        }
+
+        $modelSpaceUser = new CoreSpaceUser();
+        $data = $modelSpaceUser->getUserSpaceInfo($_SESSION["id_user"]);
```
osallou, line 133: invalid getuserspaceinfo usage ? this method returns one (and only one) user/space relation
or there is a bug in methods:

        $sql = "SELECT * FROM core_j_spaces_user WHERE id_user=?";
        return $this->runRequest($sql, array($id_user))->fetch();

should it be a fetchAll (get all spaces user is a member of)

Modules/core/Controller/CorespaceaccessController.php
gaetanHerve: It was obviously UX consideration : it seems to be more user-friendly to get all users at once, at most in the case there is not a big amount of them. But, looking at the number of requests actually required when $letter is set to "All", it is way too expensive in terms of resources.
I set it back to "A".

Modules/core/Controller/CoretilesController.php
gaetanHerve: Forgot a personal debug message. I remove it

Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -349,4 +343,59 @@ public function pendingusereditAction($id_space, $id) {
...
+            $fromName = "Platform-Manager";
+            $toAdress = $params["email"];
+            $subject = CoreTranslator::AccountCreatedSubject($lang, $spaceName);
+            $content = CoreTranslator::AccountCreatedEmail($lang, $params["login"], $params["pwd"]);
+        } else if ($origin === "accept_pending_user" || $origin === "reject_pending_user") {
+            $accepted = ($origin === "accept_pending_user") ? true : false;
+            $userModel = new CoreUser();
+            $pendingUser = $userModel->getInfo($params["id_user"]);
+            $userFullName = $pendingUser["firstname"] . " " . $pendingUser["name"];
+            $subject = CoreTranslator::JoinResponseSubject($spaceName, $lang);
```
osallou, line 390: to be coherent, params should be in same order

CoreTranslator::AccountCreatedSubject($lang, $spaceName);
vs
CoreTranslator::JoinResponseSubject($spaceName, $lang);

Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -349,4 +343,59 @@ public function pendingusereditAction($id_space, $id) {
...
+     *
+     * @param array $params required to fill sendEmail() parameters. Depends on why we want to notify user
+     * @param string $origin determines how to get sendEmail() paramters from $params
+     */
+    public function notifyUserByEmail($params, $origin) {
+        Configuration::getLogger()->debug("notifyUserByEmail", ["params" => $params]);
+        $lang = $this->getLanguage();
+        $fromName = "Platform-Manager";
+        $from = Configuration::get('smtp_from');
+        $spaceName = $params["space_name"];
```
osallou, line 378: not always set, in function pendinguserdeleteAction for example
so:
* all functions should set space_name if possible (not superadmin space for exemple)
* here, if space_name is not set, then should not be set in subject

Modules/core/Controller/CoretilesController.php
```diff
@@ -99,4 +107,99 @@ public function docAction(){
...
+     */
+    public function NotifyAdminsByEmail($params, $origin) {
+        $lang = $this->getLanguage();
+        if ($origin === "new_join_request") {
+            $spaceModel = new CoreSpace();
+            $emailSpaceManagers = $spaceModel->getEmailsSpaceManagers($params["id_space"]);
+            $mailer = new MailerSend();
+            $from = Configuration::get('smtp_from');
+            $fromName = "Platform-Manager";
+            $spaceName = $params["space_name"];
```
osallou, line 194: same remak on space_name, manage case where not in params

$spaceName = $params["space_name"] ?? "";

Modules/core/Model/CoreSpace.php
```diff
@@ -123,6 +123,16 @@ public function getSpaceIdFromName($name) {
...
     }
 
+    public function getSpaceName($id) {
+        $sql = "SELECT name FROM core_spaces WHERE id=?";
+        $req = $this->runRequest($sql, array($id));
+        if( $req->rowCount() > 0 ){
+            $tmp = $req->fetch();
+            return $tmp[0];
+        }
+        return 0;
```
osallou, line 133: should not return 0 as we expect a name (case not found)
should return null

Modules/core/Model/CoreSpaceUser.php
```diff
@@ -71,11 +71,50 @@ public function setConventionUrl($id_user, $id_space, $convention_url){
...
+            ON core_users.id = core_j_spaces_user.id_user
+            WHERE core_j_spaces_user.id_space=?
+            AND core_users.is_active=1
+            AND core_users.validated=1";
+
+        $sql .= ($active === 0)
+            ? " AND core_j_spaces_user.status=0"
+            : " AND core_j_spaces_user.status>0";
+
+        $sql .= " AND name LIKE '" . $letter . "%' ORDER BY name ASC";
```
osallou, line 116: if letter == "" , case where we want all, we should not add this LIKE condition (for a like on nothing, not efficient)

Modules/core/Model/CoreTranslator.php
```diff
@@ -109,11 +109,14 @@ public static function My_Account($lang = "") {
...
     }
     
-    public static function Account($lang = "") {
+    public static function AccountCreatedSubject($lang = "", $spaceName = "") {
+        $str = ($spaceName !== "")
+            ? ": " . $spaceName
+            : "";
         if ($lang == "fr") {
-            return "Compte utilisateur";
+            return "[pfm " . $str . "] Compte utilisateur";
```
osallou, line 117: shoud be  [pfm:  xx] xxxx  or xxxxx if spaceName is empty

Modules/core/Controller/CoreaccountController.php
```diff
@@ -69,16 +69,10 @@ public function indexAction() {
...
-                    $from = $mail_from;
-                }
-                else {
-                    $from = "support@platform-manager.com";
-                }
+                $from = Configuration::get('smtp_from');
                 $fromName = "Platform-Manager";
                 $toAdress = $form->getParameter("email");
-                $subject = CoreTranslator::Account($lang);
+                $subject = CoreTranslator::AccountCreatedSubject($lang);
```
osallou, line 75: wen user is created in a space, user is autojoined

Either:
* account creation email is sent AND a "join" mail is send prefixed by space name in subject
* account creation email only is sent, and spaceName should be prefixed in email (who created his account)

Modules/core/Controller/CoretilesController.php
```diff
@@ -99,4 +107,99 @@ public function docAction(){
...
+            "userPendingSpaceIds" => $userPendingSpaceIds,
+            "spacesUserIsAdminOf" => $spacesUserIsAdminOf
+        );
+
+        return $result;
+    }
+
+    /**
+     * 
+     * Manage actions resulting from user request to join or leave a space
```
osallou, line 154: should specify that if user is a member, then he leaves, else it is a join

Modules/core/Controller/CoretilesController.php
```diff
@@ -99,4 +107,99 @@ public function docAction(){
...
+        $modelSpaceUser = new CoreSpaceUser();
+        $isMemberOfSpace = $modelSpaceUser->exists($_SESSION["id_user"], $id_space);
+        if ($isMemberOfSpace) {
+            $modelSpaceUser = new CoreSpaceUser();
+            $modelSpaceUser->delete($_SESSION["id_user"], $id_space);
+        } else if (!$isMemberOfSpace) {
+            $spaceModel = new CoreSpace();
+            $spaceName = $spaceModel->getSpaceName($id_space);
+            $modelSpacePending = new CorePendingAccount();
+            $modelSpacePending->add($_SESSION["id_user"], $id_space);
```
osallou, line 169: you do not check here if user is already in pending

if already in pending and a new join request is done, you will have multiple entries for same user/space in db

Modules/core/Controller/CoreaccountController.php
gaetanHerve: So I modified CoreaccountController.php  but it appeared, when I tried to test it, that I can't find the way to access it by the UI. When trying to type directly the url "{host}/corecreateaccount", i'm getting a 403.
I don't know in which case it is used, but it's now meant to work with use of the new Email.php class

osallou: For self user registration?
It is disabled in ui, for routes i don t know as it does not appear in ui

gaetanHerve: Yes. The original modification I made was only because I had changed Account() into AccountCreatedSubject()  - for a semantic purpose - in CoreTranslator and checked if any files were impacted

Framework/Email.php
```diff
@@ -0,0 +1,137 @@
...
+            echo " use copy to from <br/>";
+        }
+    }
+
+    public function sendEmail($from, $fromName, $toAdress, $subject, $content, $sentCopyToFrom = false ) {
+
+        // send the email
+        $mail = new PHPMailer();
+        $mail->IsHTML(true);
+        $smtp_host = getenv('SMTP_HOST');
```
osallou, line 27: not using config smtp_host / smtp_port

Framework/Email.php
```diff
@@ -0,0 +1,137 @@
...
+require_once 'Modules/mailer/Model/MailerTranslator.php';
+use PHPMailer\PHPMailer\PHPMailer;
+
+/**
+ * Class defining methods to send an email
+ *
+ */
+class Email extends Model {
+
+    public function sendEmailSimulate($from, $fromName, $toAdress, $subject, $content, $sentCopyToFrom = true) {
```
osallou, line 13: to remove

Framework/Email.php
```diff
@@ -0,0 +1,137 @@
...
+        if ($origin === "new_join_request") {
+            $spaceModel = new CoreSpace();
+            $emailSpaceManagers = $spaceModel->getEmailsSpaceManagers($params["id_space"]);
+            $from = Configuration::get('smtp_from');
+            $from = "support@platform-manager.org";
+            $spaceName = ($params["space_name"] !== null) ? $params["space_name"] : "";
+            $fromName = "Platform-Manager";
+            $subject = CoreTranslator::JoinRequestSubject($params["space_name"], $lang);
+            $content = CoreTranslator::JoinRequestEmail($_SESSION['login'], $spaceName, $lang);
+            foreach ($emailSpaceManagers as $emailSpaceManager) {
```
osallou, line 92: sendMail supports an array of email, those will be bcc , so looping here will send X mails instead of sending 1 mail just users in bcc
Just need to set toAddress in sendEmail in expected format

PR #163, opened 2021-06-15
Pfm2 openid: add openid authentication support

osallou:
Orcid (tested with sandbox)
Should add Google auth


osallou:
already merged in pfm2...

PR #181, opened 2021-06-21
Feature issue#105

gaetanHerve:


Framework/Form.php
```diff
@@ -631,7 +639,8 @@ public function getHtml($lang = "en", $headers = true) {
                 $html .= $formHtml->hidden($this->names[$i], $this->values[$i], $required);
             }
             if ($this->types[$i] == "text") {
-                $html .= $formHtml->text($validated, $this->labels[$i], $this->names[$i], $this->values[$i], $this->enabled[$i], $required, $this->labelWidth, $this->inputWidth);
+                // #105: add readonly
+            $html .= $formHtml->text($validated, $this->labels[$i], $this->names[$i], $this->values[$i], $this->enabled[$i], $required, $this->labelWidth, $this->inputWidth, $readonlyElem);
```
osallou, line 643: readonlyElem is not initalized if type != text

Modules/core/Model/CoreUser.php
```diff
@@ -591,6 +591,26 @@ public function isLogin($login) {
...
+     * Check if an existing user, excluding actual user, already has this login
+     * 
+     * @param string $login
+     *          the login on which we are looking for existing entries in database
+     * @param string $actualUserLogin
+     *          excluded login
+     * 
+     * @return bool
+     */
+    public function isAnotherUserLogin($login, $actualUserLogin) {
```
osallou, line 605: i don't understand what is purpose here....
why checking login field ????

Modules/core/Controller/CoreusersController.php
```diff
@@ -117,7 +119,7 @@ public function editAction($id) {
         }
         $script = "";
         if ($form->check()) {
-            if ($modelUser->isLogin($this->request->getParameter('login'))) {
+            if ($modelUser->isAnotherUserLogin($this->request->getParameter('login'), $user["login"])) {
```
osallou, line 122: why calling an other fonction here?
We just want to differentiate "create new user" vs "edit existing user". If id==0, then it is a create, else an edit

osallou (changes requested):

Modules/core/Controller/CoreusersController.php
gaetanHerve: That's because the bug avoiding to validate modifications on user was linked to this function. In fact, CoreUser::isLogin($login) method returns true, and then cancels user modification, if the login already exists in database. Obviously, when editing a user (and not creating a new one), it is always true. So I choose to create a new function, working in both cases, taking into account the login of the user we want to edit. (cf CoreUser::isAnotherUserLogin($login, $actualUserLogin)). Didn't know how to name this function more clearly.

PR #182, opened 2021-06-21
Fix core pending account

gaetanHerve:


Modules/core/Controller/CoretilesController.php
```diff
@@ -157,36 +157,36 @@ public function getUserSpaces() {
      * @param int $id_space
      * @param bool $isMemberOfSpace
      */
-    public function selfJoinSpaceAction($id_space) {
+    public function selfJoinSpaceAction($space_id) {
```
osallou, line 160: should keep id_space to be coeherent with other and simply updated related route param

osallou (changes requested):

PR #187, opened 2021-06-23
Fix issue#135

gaetanHerve:
* Sending emails from Mailer Module only for members of a space works.
* Refactored email sending functions between MailerController.php and Email.php
* Left to test use of sendEmailToSpaceMembers() linked to Booking module (see Email.php line 97 on this commit).

### DIFF
--- a/Framework/Email.php
+++ b/Framework/Email.php
@@ -43,7 +43,7 @@ class Email extends Model {
         }
 
         if (is_array($toAdress)){
-            foreach($toAdress as $address){
+            foreach($toAdress as $address){                
                 $mail->addBCC($address);
             }
         } else if ( $toAdress != "" ) {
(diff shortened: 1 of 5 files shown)


Framework/Email.php
```diff
@@ -63,30 +63,87 @@ public function sendEmail($from, $fromName, $toAdress, $subject, $content, $sent
...
+                $toAddress =
+                    $this->formatAddresses($modelSpace->getEmailsSpaceActiveUsers($spaceId));
+                break;
+            case "managers":
+                $toAddress =
+                    $this->formatAddresses($modelSpace->getEmailsSpaceManagers($spaceId));
+                break;
+            default:
+                // TODO: test that! => linked to booking module. Create a dedicated function ?
+                $modelCalEntry = new BkCalendarEntry();
```
osallou, line 98: nope, Email  should not have anything related to modules but user or space models (core module)

module calling this should in this case get email addresses and call Email with a list of emails

PR #190, opened 2021-06-24
Fix issue#189 booking

gaetanHerve:


Modules/booking/View/Booking/bookday.php
```diff
@@ -88,17 +88,13 @@
 $dayNumStream = date("d", $time);
 $yearStream = date("Y", $time);
 $sufixStream = date("S", $time);
+$day_position = "";
```
osallou, line 91: Inutile

Modules/booking/View/Booking/bookday.php
```diff
@@ -88,17 +88,13 @@
...
 $day_position = date("w", $time); // 0 for sunday, 6 for saturday
-for ($p = 0 ; $p < count($day_position) ; $p++){
-	if ($day_position[$p] == 0){
-		$day_position[$p] = 7;
-	}
-}
+$day_position = ($day_position === "0") ? 7 : (int)$day_position;
```
osallou, line 93: intval($day_position)

Modules/booking/Model/BkCalendarEntry.php
```diff
@@ -241,7 +241,7 @@ public function getDefault($start_time, $end_time, $resource_id, $id_user) {
             "quantities" => "",
             "supplementaries" => "",
             "package_id" => 0,
-            "responsible_id" => $resps[0]['id'],
+            "responsible_id" => $resps['id'],
```
osallou, line 244: If resp is null, won t work

Framework/Email.php
```diff
@@ -63,30 +64,76 @@ public function sendEmail($from, $fromName, $toAdress, $subject, $content, $sent
...
             $spaceName = ($params["space_name"] !== null) ? $params["space_name"] : "";
             $fromName = "Platform-Manager";
             $subject = CoreTranslator::JoinRequestSubject($params["space_name"], $lang);
             $content = CoreTranslator::JoinRequestEmail($_SESSION['login'], $spaceName, $lang);
-            $toAdress = array();
-            foreach ($emailSpaceManagers as $emailSpaceManager) {
-                array_push($toAdress, $emailSpaceManager["email"]);
-            }
-            $this->sendEmail($from, $fromName, $toAdress, $subject, $content, false);
+            $toAddress = array();
```
osallou, line 129: Not needed, init right after

Modules/booking/Controller/BookingdefaultController.php
```diff
@@ -550,19 +551,22 @@ public function editreservationqueryAction($id_space) {
             if( !$BkUseRecurentBooking || $periodic_option == 1 ){
                 $content .= " with periodicity";
             }
-
-            //echo "send email: from " . $from . ", subject " . $subject . ", content: " . $content;
-            // send the email
-            $mailerModel = new MailerSend();
-            $mailerModel->sendEmail($from, $fromName, $toAdress, $subject, $content);
+            // NEW MAIL SENDER
+            $params = array();
```
osallou, line 555: Not needed, init rigth after

Modules/booking/Controller/BookingperiodicController.php
```diff
@@ -487,9 +487,19 @@ public function deleteAction($id_space, $id){
             //echo "send email: from " . $from . ", subject " . $subject . ", content: " . $content;
             
             // send the email
-            $mailerModel = new MailerSend();
-            $mailerModel->sendEmail($from, $space["name"], $toAdress, $subject, $content);
-
+            /* $mailerModel = new MailerSend();
+            $mailerModel->sendEmail($from, $space["name"], $toAdress, $subject, $content); */
+            // NEW MAIL SENDER
+            $params = array();
```
osallou, line 493: Not needed

Modules/booking/View/Booking/bookdayarea.php
```diff
@@ -126,11 +126,7 @@
 $yearStream = date("Y", $time);
 $sufixStream = date("S", $time);
 $day_position = date("w", $time); // 0 for sunday, 6 for saturday
-for ($p = 0 ; $p < count($day_position) ; $p++){
-	if ($day_position[$p] == 0){
-		$day_position[$p] = 7;
-	}
-}
+$day_position = ($day_position === "0") ? 7 : (int)$day_position;
```
osallou, line 129: Again intval()

Modules/core/Model/CoreConfig.php
```diff
@@ -70,9 +70,9 @@ public function isKey($key, $id_space) {
      * @param string $id_space
      */
     private function loadParams($id_space) {
-        if(CoreConfig::$params != null) {
+        /* if(CoreConfig::$params != null) {
```
osallou, line 73: $params is not a static variable?
Could explain test against null, to avoid reloading params from db if already loaded.

Linked to other space config issue, where, in addition to insert/update operation, static var should also be updated.

So not an issue to remove test, but would avoid a new db call (means that params should be loaded when accessing a space...)

Framework/Email.php
```diff
@@ -36,8 +36,10 @@ public function sendEmail($from, $fromName, $toAdress, $subject, $content, $sent
         // parse content
         $content = preg_replace("/\r\n|\r/", "<br />", $content);
         $content = trim($content);
+        Configuration::getLogger()->debug("content", ["content" => $content]);
 
         $mail->Body = $content;
+        Configuration::getLogger()->debug("content", ["content" => $content]);
```
osallou, line 42: log duplication....

Framework/Email.php
```diff
@@ -95,17 +97,20 @@ public function sendEmail($from, $fromName, $toAdress, $subject, $content, $sent
                     $this->formatAddresses($modelSpace->getEmailsSpaceManagers($spaceId));
                 break;
             default:
-            $toAddress = $params["to"];
+                try {
+                    $toAddress = $this->formatAddresses($params["to"]);
+                } catch (Exception $e) {
+                    Configuration::getLogger()->error('something went wrong getting email addresses', ['error' => $e->getMessage()]);
```
osallou, line 103: should you not return in this case? the sending will fail

Modules/booking/Controller/BookingperiodicController.php
```diff
@@ -154,15 +153,16 @@ public function editreservationqueryAction($id_space) {
...
+        $modelResp = new ClClientUser();
+        $userResps = $modelResp->getUserClientAccounts($recipient_id, $id_space);
         $foundResp = false;
         foreach($userResps as $uresp){
             if($uresp["id"] == $responsible_id){
                 $foundResp = true;
                 break;
             }
         }
+        Configuration::getLogger()->debug(["TEST"], ["after new EcResponsible" => "var2"]);
```
osallou, line 165: for your debug, not a good debug line here

Modules/booking/Controller/BookingperiodicController.php
```diff
@@ -457,46 +450,27 @@ private function isNew($param) {
     }
 
     public function deleteAction($id_space, $id){
+        $sendEmail = $this->request->getParameter("sendmail");
         
-        $sendmail = $this->request->getParameter("sendmail");
-        
-        if( $sendmail == 1){
+        if( $sendEmail == 1){
```
osallou, line 455: does it works with php8 (sendmail will be a string no?)

Modules/mailer/Controller/MailerController.php
```diff
@@ -77,11 +77,17 @@ public function sendAction($id_space) {
         $subject = $this->request->getParameter("subject");
         $content = $this->request->getParameter("content");
 
-        $modelConfig = new CoreConfig();
-        $mailerSetCopyToFrom = $modelConfig->getParamSpace("MailerSetCopyToFrom", $id_space);
-        $mailerSetCopyToFromBool = false;
-        if($mailerSetCopyToFrom == 1){
-            $mailerSetCopyToFromBool = true;
+        if (!in_array($to, array("all", "managers"))) {
+            $toEx = explode("_", $to);
```
osallou, line 81: explode by underscore ?  the $to string contains a list of mails with underscore as delimiter ?
weird ? how is it set ? usually, we have comma separated list in this case

Modules/seek/Controller/SeekbookingController.php
```diff
@@ -471,37 +465,30 @@ private function isNew($param) {
 
     public function deleteAction($id_space, $id){
         
-        $sendmail = $this->request->getParameter("sendmail");
+        $sendEmail = $this->request->getParameter("sendmail");
         
-        if( $sendmail == 1){
+        if( $sendEmail == 1){
```
osallou, line 470: same comment as above for comparison , sendmail = "1" no ? should work, but bad practive to compare string to integer

Modules/booking/Controller/BookingdefaultController.php
```diff
@@ -293,31 +284,21 @@ public function editreservationqueryAction($id_space) {
             }
         } else {
             $periodicEndDate = CoreTranslator::dateToEn($this->request->getParameter("periodic_enddate"), $lang);
-            $periodicEndDateArray = explode("-", $periodicEndDate);
-            $last_start_time = mktime($hour_startH, $hour_startM, 0, $periodicEndDateArray[1], $periodicEndDateArray[2], $periodicEndDateArray[0]);
-
+            $periodicEndDateArray = explode("/", $periodicEndDate);
```
osallou, line 287: separator is different here, dateToEn returns with "-" or "/" ????  how is it done with other dateToEn calls?

Modules/booking/Controller/BookingdefaultController.php
```diff
@@ -831,75 +767,36 @@ public function deleteperiodAction($id_space, $id_period){
...
-            $from = Configuration::get('smtp_from');
-            $fromName = "Platform-Manager";
+            $toAddress = $modelCalEntry->getEmailsBookerResource($id_resource);
 
-            $toAdress = $modelResource->getResourcesManagersEmails($id_resource);
-            $subject = "[". $space["name"] ."]" . $resourceName . " has been freed";
+            $modelConfig = new CoreConfig();
+            $sendMailResponsibles = $modelConfig->getParamSpace("BkBookingMailingAdmins", $id_space);
+            
+            if ($sendMailResponsibles > 0) {
```
osallou, line 783: php8 comparaison string/int again, works?

osallou (changes requested):

Modules/booking/Controller/BookingdefaultController.php
osallou: dateToEn, if lang=en, will return date, but if lang=fr will return YYYY-MM-DD

so i suppose you want to explode original date, not translated date (I am not sure with parameter gives according to lang)

    $periodicEndDate = CoreTranslator::dateToEn($this->request->getParameter("periodic_enddate"), $lang);
    $periodicEndDateArray = explode("/", $this->request->getParameter("periodic_enddate"));

osallou: in addreservationaction , dates taken as:

    $('.form_date_fr').datetimepicker({
                    language: 'fr',
                    weekStart: 1,
                    todayBtn: 1,
                    autoclose: 1,
                    todayHighlight: 1,
                    startView: 2,
                    minView: 2,
                    forceParse: 0,
                    format: 'dd/mm/yyyy'
                });
                $('.form_date_en').datetimepicker({
                    language: 'us',
                    weekStart: 1,
                    todayBtn: 1,
                    autoclose: 1,
                    todayHighlight: 1,
                    startView: 2,
                    minView: 2,
                    forceParse: 0,
                    format: 'yyyy-mm-dd'
                });


so different format according to lang, though in code only form_date_fr is set... so we can suppose we always receive (at least here) dd/mm/yyyy so dateToEn will return YYY-MM-DD

In english, will also have YYY-MM-DD, so should indeed keep original   $periodicEndDateArray = explode("-", $periodicEndDate)

Modules/booking/Controller/BookingperiodicController.php
gaetanHerve: Yes sendemail is a string and this works in php8.

Modules/mailer/Controller/MailerController.php
gaetanHerve: I kept the mechanism Sylvain was using. In fact, the input select where the user selects to which group he wants to send the email returns values like "a_1" for an area which id is ,1 or "r_2" for a resource which id is 2.
He could also have extracted the 1st char (with a substr()), but not sure it would have been more sensible. explode() also appears to be pretty fast.

Modules/booking/Controller/BookingdefaultController.php
gaetanHerve: Thanks for having brought that to my attention. I went too fast on this. The original code wasn't working for me. I logged the date and it was including "/", not "-". I gonna take language into account to make it work in both case

gaetanHerve: It appeared that date format isn't always linked to $lang (at least in bookingdefaultController) and, consequently, if $lang="en" and date format is dd/mm/yyyy, dateToEn() in CoreTranslator.php will return a date in french format and an error will be risen.
It is possible that this case stands in multiple places, so, for a short-term efficient fix, I replaced in dateToEn($date, $lang) ```if ($lang == "fr")``` by ```if ($lang == "fr" || str_contains($date, "/"))```

gaetanHerve: see issue #200

Modules/booking/Controller/BookingdefaultController.php
gaetanHerve: sill works. I gonna try to explicitly make the conversions when I see that

PR #205, opened 2021-07-05
super admin can no longer delete a user active or pending in a space

gaetanHerve:
when space admin clicks on delete button for a user on page {host}/coreusers then confirms:
* if user is neither active nor pending in a space, he is deleted
* else, a flash message alerts that user deletion is not allowed

### DIFF
--- a/Modules/core/Model/CoreSpaceUser.php
+++ b/Modules/core/Model/CoreSpaceUser.php
@@ -54,6 +54,20 @@ class CoreSpaceUser extends Model {
         return false;
     }
     
+    /**
+     * 
+     * Returns true if user is an active member of any space
+     * 
+     * @param int $id_user
+     * 
+     * @return bool
+     */
+    public function isActiveMemberOfAnySpace($id_user){
+        $sql = "SELECT id FROM core_j_spaces_user WHERE id_user=?";
+        $req = $this->runRequest($sql, array($id_user));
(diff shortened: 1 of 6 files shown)


Modules/core/Model/CoreSpaceUser.php
```diff
@@ -54,6 +54,20 @@ public function exists($id_user, $id_space){
...
     
+    /**
+     * 
+     * Returns true if user is an active member of any space
+     * 
+     * @param int $id_user
+     * 
+     * @return bool
+     */
+    public function isActiveMemberOfAnySpace($id_user){
```
osallou, line 65: There is already a method to get user role, if result is null, then there is no role.
Should be enough rather than creating a new method for that

Modules/core/View/Coreusers/indexAction.php
```diff
@@ -14,6 +14,18 @@
         </div>
     </div>
     <div>À utiliser avec précautions !!! La suppression d'utilisateurs ne supprime pas les données associées. N'utiliser qu'en cas de doublon ou de cas exotiques.</div>
+    <div class="col-md-6">
+        <?php if (isset($_SESSION["message"])) { ?>
+            <div class="alert alert-danger alert-dismissible">
+                <?php echo $_SESSION["message"] ?>
```
osallou, line 20: need to unset message else will remain after that

PR #215, opened 2021-07-09
user can re-join a space he was once member of

gaetanHerve:


osallou:
a bit complex... and for a user that left space, getting validated=1 looks strange...

gaetanHerve:
Yes. Maybe we're missing a specific attribute in object / column in database...

gaetanHerve:
Maybe let's keep it that way for now, so we can release

PR #216, opened 2021-07-09
refactor space user delete to keep only one function

osallou:


Modules/core/Model/CoreSpace.php
```diff
@@ -515,8 +516,12 @@ public function setAdmins($id, $id_admins) {
                 }
             }
             if (!$found) {
-                $sql = "DELETE FROM core_j_spaces_user WHERE id_space=? AND id_user=? AND status=?";
-                $this->runRequest($sql, array($id, $aadm["id_user"], CoreSpace::$ADMIN));
+                $m = new CoreSpaceUser();
+                $m->delete($aadm["id_user"], $id, CoreSpace::$ADMIN);
+                $modelSpacePending = new CorePendingAccount();
+                $modelSpacePending->updateWhenUnjoin($id_user, $id_space);
```
gaetanHerve, line 523: Does it work if this admin doesn't exist in core_pending_account for this space (if it was created before new release) ?

Issue #218, opened 2021-07-09
Allow user to define default language


osallou:
by default, takes navigator language

user should be able to define his default language in user preferences (via CoreUserSettings)

PR #219, opened 2021-07-13
add language setting in user settings

gaetanHerve:
* add function languageeditAction() in CoreusersController
* user can now set default language 
* closes issue #218

### DIFF
--- a/Framework/Router.php
+++ b/Framework/Router.php
@@ -51,8 +51,8 @@ class Router {
     }
 
     private function route($request) {
-
         $modulesNames = Configuration::get("modules");
+        $modulesNames = is_array($modulesNames) ? $modulesNames : [$modulesNames];
         foreach ($modulesNames as $moduleName) {
             // get the routing class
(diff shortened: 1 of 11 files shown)


Framework/Router.php
```diff
@@ -51,8 +51,8 @@ private function call($target, $args, $request) {
     }
 
     private function route($request) {
-
         $modulesNames = Configuration::get("modules");
+        $modulesNames = is_array($modulesNames) ? $modulesNames : [$modulesNames];
```
osallou, line 55: Ok, but why?
Modules are defined in config file as an array

gaetanHerve: That's a security. Documentation indicates that Configuration::get() returns a string. In some similar cases in the framework, I encountered functions returning a string for real instead of an array when there was only one element.
But you're right, in this case I doubt it to be useful.

osallou: Get doesn t always return a string, it returns result of parse ini.
For elements like:

modules[] it returns an array.
This is a doc error

PR #226, opened 2021-07-22
Fix 225 bugs from security

gaetanHerve:
Some of the fixes concern bugs coming from typos in fix_196_security. But I also, passing by, corrected some difficult to isolate small problems with the idea of just making it globally work after merge. Some issues remain and I have not tested all modules from A to Z. This already fixes a certain amounts of bugs in modules services, antibodies and statistics.

### DIFF
--- a/Modules/clients/Form/AddressForm.php
+++ b/Modules/clients/Form/AddressForm.php
@@ -44,6 +44,7 @@ class AddressForm extends FormGenerator{
     }
    
     public function save(){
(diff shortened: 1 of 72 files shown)


Modules/clients/Form/AddressForm.php
```diff
@@ -48,18 +45,17 @@ public function render(){
    
     public function save(){
         $model = new ClAddress();
-        $id = $model->set(
-                $this->request->getParameter("id"), 
+        return $model->set(
+                $this->id_space,
```
osallou, line 49: Id space is set at  form creation?

PR #235, opened 2021-07-27
replace use of se_service_types by use of a static array, closes #227

### LINKED ISSUE #227: [pfm2][Services][Fix security] Should services types be common to all spaces?
Like other model classes, SeServiceType.php was modified in PR fix_security_196 in order to integrate space ids.
But it appears that the data is not thought to be created, updated nor deleted by users (no interface for that). Sylvain only planned to have 6 fixed types so that services can use it.

Needs to be solved before upgrade 2.1.

```
public function createDefault($id_space) {
        $this->add($id_space, "Quantity", "Quantité");
        $this->add($id_space, "Time minutes", "Temps en minutes");
        $this->add($id_space, "Time hours", "Temps en heures");
        $this->add($id_space, "Price", "Prix");
        $this->add($id_space, "Half day", "Demi journée");
        $this->add($id_space, "Journée", "Journée");
    }
```
Unless we want to make it editable by users, I think we should remove the $id_space parameter from this class and add lines in db at creation (_cf._ code block above) only if it doesn't exist. I'll keep that opened waiting for further advice.

gaetanHerve:


osallou:
The problem i see with using static array is with existing db.

We have no certainty that service types currently in db are in same order from 1 to x. And existing services refer to a service type id, not its name. So we may have a servuce refering setype 3, but 3 not being third setype of your array...

By the way, when requesting by an id, you take type from setype[id] , but existing references in db reference in autoincr index, so first servicetype has id=1
In array, indice starts at 0.

I think we should keep this way in db for compatibility.... :-( what you propose is fine for a fresh new code, but may not work with amexisting data.

But you should anyway remove space id ref for setype.

gaetanHerve:
Ah, yeah, I messed up with the index. Thanks Olivier for your remark.
Looking to the dump of prod database I think there is actually no problem to replace se_service_types table by a static array: 

```
--
-- Dumping data for table `se_service_types`
--

LOCK TABLES `se_service_types` WRITE;
/*!40000 ALTER TABLE `se_service_types` DISABLE KEYS */;
INSERT INTO `se_service_types` VALUES (1,'Quantity','Quantité'),(2,'Time minutes','Temps en minutes'),(3,'Time hours','Temps en heures'),(4,'Price','Prix'),(5,'Half day','Demi journée'),(6,'Journée','Journée');
/*!40000 ALTER TABLE `se_service_types` ENABLE KEYS */;
UNLOCK TABLES;
```

Once dealt with my array index, I can't figure out what couldn't work correctly...
User interface not giving access to adding data to this table, there wont be any difference between this dump and db state for the future upgrade integrating this fix...

I'm gonna update this PR to fix the index question and let you decide about merging.

osallou:
In our database, maybe, but others?
Else on upgrade we should check types indices and if different, update references

Modules/services/Model/SeServiceType.php
```diff
@@ -9,6 +9,61 @@
...
+                    if (!in_array($serviceId["id"], $modifiedServiceIds)) {
+                        array_push($serviceIds, $serviceId["id"]);
+                    }
+                }
+                //then update
+                foreach ($serviceIds as $serviceId) {
+                    $sql = "UPDATE se_services SET type_id=? WHERE id=?";
+                    // get index by name in $serviceTypesArray
+                    $typeIndex = array_search(ServicesTranslator::serviceTypes($typesData[$i]["name"], "en"), $serviceTypesArray);
+                    Configuration::getLogger()->info("[TEST][UPDATING TYPES]", ["modified type id" => $serviceId]);
```
osallou, line 57: Test log to remove

Modules/services/Model/SeServiceType.php
```diff
@@ -9,6 +9,61 @@
...
+        // extract service types names and ids from database
+        $sql = "SELECT `id`, `name` FROM se_service_types";
+        $typesData = $this->runRequest($sql, array())->fetchAll();
+        $tmpTypes = [];
+        $modifiedServiceIds = [];
+        $serviceTypesArray = self::$serviceTypes;
+
+        for ($i=0; $i<6; $i++) {
+            // format data get from database to match $serviceTypes format for further comparisons
+            $tmpTypes[$typesData[$i]["id"]] = $typesData[$i]["name"];
```
osallou, line 36: Name or local_name?

gaetanHerve: Name. I replaced local_name values by a translation function in servicesTranslator.php. More efficient and, for an upgrade, it allows to easily find out what is the type, whether it was set as local_name or name in db.

gaetanHerve: So here, once $tmpTypes filled, I can compare its values (names from db) with (values from static array $serviceTypes or translated values from static array $serviceTypes). So we get the right corresponding indexes. Then translation is used, as usual, to adapt displaying to the user's language settings

Issue #260, opened 2021-08-17
Invalid type for year


osallou:
TypeError: mktime(): Argument #6 ($year) must be of type ?int, string given File "/Modules/booking/Controller/BookingController.php", line 396, in mktime $dateBegin = mktime(0, 0, 0, $dateArray[1], $dateArray[2], $dateArray[0]); File "/Modules/booking/Controller/BookingController.php", line 396, in BookingController::dayareaAction $dateBegin = mktime(0, 0, 0, $dateArray[1], $dateArray[2], $dateArray[0]); File "/Framework/Controller.php", line 59, in Controller::runAction call_user_func_array(array($this, $actionName), $args); File "/Modules/core/Controller/CoresecureController.php", line 120, in CoresecureController::runAction parent::runAction($module, $action, $args); File "/Framework/Router.php", line 149, in Router::runAction $controller->runAction($urlInfo["pathInfo"]["module"], $action, $args); ... (2 additional frame(s) were not displayed) 

Request

URLhttp://pfm.genouest.org/bookingdayarea/1
Method GET
Query path=bookingdayarea%2F1

gaetanHerve:
In several cases, mktime() is called with invalid arguments. It appears it often has to deal with the absence of default values. cf #297

Issue #259, opened 2021-08-16
Division by zero


osallou:
Error http://pfm.genouest.org/bookingdayarea/20

DivisionByZeroError: Division by zero File "/Modules/booking/View/Booking/bookdayarea.php", line 12, in include $dayWidth = 100/count($resourcesBase); File "/Framework/View.php", line 49, in View::generate include ($this->file); File "/Framework/Controller.php", line 114, in Controller::render $view->generate($dataView); File "/Modules/booking/Controller/BookingController.php", line 439, in BookingController::dayareaAction ), "bookdayarea"); File "/Framework/Controller.php", line 59, in Controller::runAction call_user_func_array(array($this, $actionName), $args); ... (4 additional frame(s) were not displayed)

Seems in some cases we can have 0, should handle use case ir use default

PR #263, opened 2021-08-25
add error messages to booking UI when areas and/or resources are not …

gaetanHerve:
It was linked to the calculation of column width in "dayarea" screen in booking module. In fact, width was equal to 100/nb resources. Obviously, If there is no resource in database, it triggers a bug.

We could have forbidden access to booking, but I thought it would be more user friendly to let booking interfaces display (with booking actions unavailable when no resources or areas exist in db).


osallou:
Closes #259

gaetanHerve:
Closes #260

Issue #261, opened 2021-08-19
Check permissions on pendinguserdelete


osallou:
* if exception instance of pfmauthexception, do not send to sentry
* check permissions to access pendinguserdelete
  * allowed for space managers only
  * if not allowed, delete button should not appear

PfmAuthException: Error 403: Permission denied #6 /Modules/core/Controller/CoresecureController.php(146): CoresecureController::checkAuthorization #5 /Modules/core/Controller/CorespaceaccessController.php(369): CorespaceaccessController::pendinguserdeleteAction #4 /Framework/Controller.php(59): Controller::runAction #3 /Modules/core/Controller/CoresecureController.php(120): CoresecureController::runAction #2 /Framework/Router.php(149): Router::runAction #1 /Framework/Router.php(122): Router::routerRequest #0 /index.php(20): null

Formattedcurl

GET/corespacependinguserdelete/6/30

pfm.genouest.org

Query String

path

corespacependinguserdelete/6/30

osallou:
pendinguserdeleteAction checks user is admin, it should check that user is at least manager in related space and pendinguser is in space

PR #264, opened 2021-08-25
Fix 261 permissions on pendinguserdelete

gaetanHerve:
- from sentry is linked to production environment, not dev, removal of PfmauthExceptions from exceptions send to sentry is not tested
- For purposes of coherence only space admins (and, of course, super admins) are allowed to delete pending users. "Manager" role is not enough. Same authorizations are already applied to affect role to pending users and access users module within a space.


gaetanHerve:
Closes #261

osallou (changes requested):

Issue #269, opened 2021-09-01
[Com Module] Setting Com as space home page forbid access to other modules


gaetanHerve:
When setting Com as "space home page" - in Com configuration - user has no longer access to the rest of the space. There is no way, except by typing directly an url, to access the other modules. Space home page url redirects to {host}/comhome/{id_space}.
We should add a button to access the usual interface or add the modules under the com section.

PR #272, opened 2021-09-03
Fix com home page option issue#269

gaetanHerve:


gaetanHerve:
Closes #269

osallou:
Would be nice to avoid jquery (dep quite old etc..), could use vuejs as in a few other views for interactive display.

Instead of jquery post, could also use fetch (from javascript directly, can look at helpdesk for examples). By the way, a GET would be more coherent than a post to get list of news.

At last, no need to create specific api routes. You can use a controller and in render data field add your news. When sending your GET request, just specify a json content type in accept headers, in this case render methid will return content of data structure instead of displaying page

In helpdesk, you can look at the way list of tickets is fetched

Modules/core/Controller/CorespaceController.php
```diff
@@ -57,7 +57,10 @@ public function viewAction($id_space) {
 
         $modelConfig = new CoreConfig();
         $space_home_page = $modelConfig->getParamSpace('space_home_page', $id_space);
-        if ($space_home_page != "") {
+
+        $showCom = ($space_home_page == "comhome");
+
+        if ($space_home_page != "" && !$showCom) {
```
osallou, line 63: if no space home page defined, it defaults  to space id. 


Following redirect will redirect in this case to spaceid/spaceid

gaetanHerve:
> Would be nice to avoid jquery (dep quite old etc..), could use vuejs as in a few other views for interactive display.
> 
> Instead of jquery post, could also use fetch (from javascript directly, can look at helpdesk for examples). By the way, a GET would be more coherent than a post to get list of news.
> 
> At last, no need to create specific api routes. You can use a controller and in render data field add your news. When sending your GET request, just specify a json content type in accept headers, in this case render methid will return content of data structure instead of displaying page
> 
> In helpdesk, you can look at the way list of tickets is fetched

I'm gonna create a new PR taking that into account so we can already use this one, which is working.

Modules/core/Controller/CorespaceController.php
osallou: $space_home_page = $modelConfig->getParamSpace('space_home_page', $id_space)  will return $id_space if space_home_page has never been defined...

gaetanHerve: I'm probably missing something here... 
On this PR, CoreConfig::getParamSpace():

```
/**
     * Get a parameter
     * @param string $key
     * @return string value
     */
    public function getParamSpace($key, $id_space) {
        $this->loadParams($id_space);
        if(!isset(self::$params[$id_space])) {
            return "";
        }
        if(!isset(self::$params[$id_space][$key])) {
            return "";
        }
        return self::$params[$id_space][$key];
    }
```

And, in db, the keyname "space_home_page" doesn't exist for a space if it has never been defined. In this case, getParamSpace() returns an empty string ($this->loadParams() just setting self::$params as an array [$id_space][$key]). No ?

osallou: ah ok, my fault, I was thinking about an other option to get config params where second param is default value...

gaetanHerve: Ah, ok :). No problem.

PR #282, opened 2021-09-09
add email duplicates checks in core users and space users forms

### LINKED ISSUE #278: [Users duplication] forbid usage of same email address for multiple users
forbidding usage of same email address for multiple users will avoid duplication. We have to alert user at account creation

gaetanHerve:
- closes #278 
- merged into form_data_staging - MR #286

### DIFF
--- a/Modules/core/Controller/CorespaceaccessController.php
+++ b/Modules/core/Controller/CorespaceaccessController.php
@@ -202,12 +202,19 @@ class CorespaceaccessController extends CoresecureController {
         $form->setValidationButton(CoreTranslator::Ok($lang), "corespaceaccessuseradd/".$id_space);
 
         if ($form->check()) {
-
             $modelCoreUser = new CoreUser();
+            $canEditUser = true;
+            if ($modelCoreUser->isLogin($this->request->getParameter('login'))) {
+                $canEditUser = false;
+                throw new PfmException(CoreTranslator::LoginAlreadyExists($lang), 403);
+            }
+            if($modelCoreUser->isEmail($form->getParameter("email"))) {
+                // if email alreday exists, warn user
+                $canEditUser = false;
+                $_SESSION["flash"] = CoreTranslator::EmailAlreadyExists($lang);
+            }
 
-            if ($modelCoreUser->isLogin($form->getParameter("login"))) {
-                $_SESSION["message"] = CoreTranslator::Error($lang) . ":" . CoreTranslator::LoginAlreadyExists($lang);
-            } else {
+            if ($canEditUser) {
                 $pwd = $modelCoreUser->generateRandomPassword();
 
                 $id_user = $modelCoreUser->createAccount(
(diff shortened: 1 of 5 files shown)


Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -202,12 +202,19 @@ public function useraddAction($id_space){
         $form->setValidationButton(CoreTranslator::Ok($lang), "corespaceaccessuseradd/".$id_space);
 
         if ($form->check()) {
-
             $modelCoreUser = new CoreUser();
+            $canEditUser = true;
+            if ($modelCoreUser->isLogin($this->request->getParameter('login'))) {
+                $canEditUser = false;
+                throw new PfmException(CoreTranslator::LoginAlreadyExists($lang), 403);
```
osallou, line 209: why throwing an exception here but not in email case?
both should raise an exception or display flash error

Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -202,12 +202,19 @@ public function useraddAction($id_space){
...
+            }
+            if($modelCoreUser->isEmail($form->getParameter("email"))) {
+                // if email alreday exists, warn user
+                $canEditUser = false;
+                $_SESSION["flash"] = CoreTranslator::EmailAlreadyExists($lang);
```
osallou, line 214: flashClass should be error

Modules/core/Controller/CoreusersController.php
```diff
@@ -125,14 +125,36 @@ public function editAction($id) {
         }
         $script = "";
         if ($form->check()) {
-            if (!$id && $modelUser->isLogin($this->request->getParameter('login'))) {
-                throw new PfmException(CoreTranslator::LoginAlreadyExists($lang), 403);
+            $canEditUser = true;
+            if (!$id) {
+                if ($modelUser->isLogin($this->request->getParameter('login'))) {
+                    $canEditUser = false;
+                    throw new PfmException(CoreTranslator::LoginAlreadyExists($lang), 403);
```
osallou, line 132: same remark for exception

Modules/core/Controller/CoreusersController.php
```diff
@@ -125,14 +125,36 @@ public function editAction($id) {
...
+                }
+                if($modelUser->isEmail($form->getParameter("email"))) {
+                    // if email alreday exists, warn user
+                    $canEditUser = false;
+                    $_SESSION["flash"] = CoreTranslator::EmailAlreadyExists($lang);
+                }
             } else {
+                $emailUsedByAnother = $modelUser->isEmailWithFilter($form->getParameter("email"), $user["email"]);
```
osallou, line 140: why a filter ?
if param email == user['email'], that's fine (no change)
if different, should be used nowhere so isEmail should be enough

Modules/core/Controller/CorespaceaccessController.php
gaetanHerve: It was a PfmException at first for login. I think you put it there. But I prefer flash messages, more user friendly. I push back with a flash message

Modules/core/Controller/CoreusersController.php
gaetanHerve: ah, yes. It's simpler

Modules/core/Controller/CorespaceaccessController.php
gaetanHerve: ok, I add that

PR #286, opened 2021-09-14
Form data staging

gaetanHerve:
Not working for now...

Use of a js script to check unicity of emails and logins to warn the user before form submission. Avoid to lose filled inputs when informations not accepted.

### DIFF
--- a/Framework/FormHtml.php
+++ b/Framework/FormHtml.php
@@ -394,7 +394,7 @@ class FormHtml {
         $html .= "<input class=\"form-control\" type=\"email\" id=\"" . $name . "\" name=\"" . $name . "\"";
         $html .= " value=\"" . $value . "\"" . $required;
         if ($checkUnicity) {
-            $html .= "onfocusout=\"checkEmailUnicity() \"";
+            $html .= "onblur=\"checkEmailUnicity() \"";
         }
         $html .= "/>";
         $html .= "</div>";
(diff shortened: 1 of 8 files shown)


Modules/core/Controller/CoreusersController.php
```diff
@@ -165,7 +199,7 @@ protected function editQuery($form, $modelUser, $lang) {
             $pwd = $form->getParameter("pwd");
             $pwdconfirm = $form->getParameter("pwdconfirm");
             if ($pwd != $pwdconfirm) {
-                throw new PfmException(CoreTranslator::TheTwoPasswordAreDifferent($lang), 403);
+                $_SESSION['flash'] = CoreTranslator::TheTwoPasswordAreDifferent($lang);
```
osallou, line 202: should not add user in this case

Framework/Router.php
```diff
@@ -128,6 +128,7 @@ public function routerRequest() {
                 }
 
                 $urlInfo = $this->getUrlData($request);
+                Configuration::getLogger()->debug("[TEST][ROUTER]", ["urlInfo" => $urlInfo]);
```
osallou, line 131: debug log to remove

Framework/FormHtml.php
```diff
@@ -723,4 +736,12 @@ static public function ajaxScript($formId, $validationURL) {
         return str_replace("validationurl", $validationURL, $string1);
     }
 
+    /**
+     * 
+     * @return type
+     */
+    static public function checkUnicityScript() {
+        // return file_get_contents("Framework/checkUnicity_script.php");
```
osallou, line 744: ?? empty

osallou (changes requested):

PR #297, opened 2021-09-28
add cast to int on mktime() arguments

gaetanHerve:
haven't found from where else the error could come from... Haven't managed to reproduce bug on local environnement


gaetanHerve:
We could merge this since we merged the other relative to mktime errors fixers...

PR #315, opened 2021-10-07
fix sql error in booking edit

gaetanHerve:
this error is not a blocking error...


Modules/resources/Model/ReEvent.php
```diff
@@ -36,7 +36,7 @@ public function getLastStateColor($id_space, $id_resource){
         $sql = "SELECT re_state.color as color"
                 . " FROM re_event "
                 . " INNER JOIN re_state ON re_event.id_state = re_state.id"
-                . " WHERE id_resource=? AND id_space=? AND deleted=0 ORDER BY date DESC;";
+                . " WHERE id_resource=? AND re_event.id_space=? AND deleted=0 ORDER BY date DESC;";
```
osallou, line 39: not needed also on deleted?

PR #343, opened 2021-10-14
Add client infos to invoices template

gaetanHerve:


Modules/quote/Model/Quote.php
```diff
@@ -51,10 +51,11 @@ public function getAllInfo($id_space, $id) {
 
             $data["recipient"] = $modelUser->getUserFUllName($data["id_user"]);
             $resps = $modelUserClient->getUserClientAccounts($data["id_user"], $id_space);
-            if (count($resps) > 0) {
+            if ($resps && !empty($resps)) {
                 $data["address"] = $modelClient->getAddressInvoice($id_space, $resps[0]["id"]);
```
osallou, line 55: On ne l a pas ds client?

gaetanHerve: No, the sql request is a join with cl_addresses table

osallou (commented):
le changelog ? ....

PR #371, opened 2021-10-27
Selfregistration improvement

### LINKED ISSUE #362: [Registration] self registration improvement
When a user self registers, by this way asks to join an existing space, space admins should be given clues about user's origin.
So, we should add fields to self registration form:
-  "Orgnisation,  tutelle ou entreprise"
- "Equipe, labo ou service"

We should not propose, as the user has chosen a space to which he will request join, a list of structures linked to this space.This would cause a discretion problem.
We could suggest a default login (firstname first letter + last name)
Unlike in users creation forms (for space admins or superadmins), I believe there is no unicity checks on login and email in the self registration form. Need to add it.

Space admins from Rennes instance want it to be activated.

gaetanHerve:
closes #362
Warning: changes on coresecureController.php to allow call to coreUsers from coreaccountcreation page for unicity checking.
What about SQL injection in PFM forms in general ? Does PDO give enough security in our case ?

### DIFF
--- a/Framework/Form.php
+++ b/Framework/Form.php
@@ -85,6 +85,7 @@ class Form {
         $this->isFormAdd = false;
         $this->externalButtons = array();
         $this->useAjax = $useAjax;
+        $this->suggestLogin = false;
 
         $this->parseRequest = false;
         $formID = $request->getParameterNoException("formid");
@@ -214,7 +215,6 @@ class Form {
         $this->submitOnChange[] = false;
         $this->readonly[] = false;
         $this->checkUnicity[] = false;
-        $this->suggestLogin[] = false;
     }
 
     /**
@@ -235,7 +235,6 @@ class Form {
         $this->submitOnChange[] = false;
         $this->readonly[] = false;
         $this->checkUnicity[] = false;
-        $this->suggestLogin[] = false;
     }
 
     /**
@@ -256,7 +255,6 @@ class Form {
         $this->submitOnChange[] = false;
         $this->readonly[] = false;
         $this->checkUnicity[] = false;
-        $this->suggestLogin[] = false;
     }
 
     /**
@@ -279,7 +277,6 @@ class Form {
         $this->submitOnChange[] = false;
         $this->readonly[] = false;
         $this->checkUnicity[] = false;
-        $this->suggestLogin[] = false;
     }
 
     /**
@@ -303,7 +300,6 @@ class Form {
         $this->submitOnChange[] = false;
         $this->readonly[] = false;
         $this->checkUnicity[] = false;
-        $this->suggestLogin[] = false;
     }
 
     /**
@@ -325,7 +321,6 @@ class Form {
         $this->setValue($name, $url);
         $this->readonly[] = false;
         $this->checkUnicity[] = false;
-        $this->suggestLogin[] = false;
     }
 
     /**
@@ -350,7 +345,9 @@ class Form {
         $this->submitOnChange[] = false;
         $this->readonly[] = $readonly;
         $this->checkUnicity[] = $checkUnicity;
-        $this->suggestLogin[] = $suggestLogin;
+        if ($suggestLogin) {
+            $this->suggestLogin = true;
+        }
     }
 
     /**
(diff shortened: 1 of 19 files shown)


Modules/users/Controller/UseraccountController.php
```diff
@@ -54,8 +54,8 @@ public function indexAction() {
         $form->addText("firstname", CoreTranslator::Firstname($lang), true, $userCore["firstname"]);
         $form->addText("name", CoreTranslator::Name($lang), true, $userCore["name"]);
         $form->addEmail("email", CoreTranslator::Email($lang), true, $userCore["email"]);
-        
-        $form->addText("unit", UsersTranslator::Unit($lang), false, $userInfo["unit"] ?? "");
+        $form->addText("organization", CoreTranslator::Organization($lang), false, $userInfo["organization"] ?? "");
```
osallou, line 57: should not it be mandatory?

osallou (changes requested):
and I do not see where new org info is displayed after that (beside email)

should appear in pending I think, as well as user list in spaceadmin

Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -118,11 +119,16 @@ public function indexAction($id_space, $letter = "A", $active = "") {
...
 
         // get user list
+        
         $usersArray = array();
         $isActive = ($active === "active") ? 1 : 0;
         $modelSpaceUser = new CoreSpaceUser();
+        $modelUsersInfo = new UsersInfo();
         $users = $modelSpaceUser->getUsersOfSpaceByLetter($id_space, $letter, $isActive);
         foreach ($users as $user) {
+            $userSupInfos = $modelUsersInfo->get($user['id']);
```
osallou, line 129: not cool here, one request per user....

getUsersIfSpace... should return info via a join (or using an other request with the join)

Modules/core/Controller/CorespaceaccessController.php
```diff
@@ -351,14 +359,18 @@ public function pendingusersAction($id_space) {
...
         $modelSpacePending = new CorePendingAccount();
         $modelUser = new CoreUser();
+        $modelUserInfos = new UsersInfo();
 
         $pendingUsers = [];
         $data = $modelSpacePending->getPendingForSpace($id_space);
         for ($i = 0; $i < count($data); $i++) {
             $pendingUsers[] = $data[$i];
             $userInfos = $modelUser->getInfo($data[$i]['id_user']);
+            $userSupInfos = $modelUserInfos->get($data[$i]['id_user']);
```
osallou, line 369: again a request per user

Framework/Form.php
```diff
@@ -343,6 +345,9 @@ public function addText($name, $label, $isMandatory = false, $value = "", $enabl
         $this->submitOnChange[] = false;
         $this->readonly[] = $readonly;
         $this->checkUnicity[] = $checkUnicity;
+        if ($suggestLogin) {
+            $this->suggestLogin = true;
```
osallou, line 349: not suggestLogin[] ?

Framework/Form.php
```diff
@@ -742,9 +747,13 @@ public function getHtml($lang = "en", $headers = true) {
             $html .= $formHtml->textAreaScript();
         }
 
-        if ($this->checkUnicity == true) {
+        if (in_array(true, $this->checkUnicity)) {
             $html .= $formHtml->checkUnicityScript();
         }
+        
+        if ($this->suggestLogin) {
```
osallou, line 754: why not in array with in_array too?

Modules/core/Controller/CoreusersController.php
```diff
@@ -213,18 +219,30 @@ protected function editQuery($form, $modelUser, $lang) {
...
+                    "",
+                    $form->getParameter("unit"), 
+                    $form->getParameter("organization")
+                );
             }
         } else {
             $modelUser->edit(
                 $id,
-                $form->getParameter("login"),
+                $modelUser->getInfo($id)["login"],
```
osallou, line 232: if getinfo null => exception

osallou (approved):

PR #384, opened 2021-11-08
Feature welcomepage

osallou:
Change welcome page (coretiles) to be a simple page with:

* a welcome text, extracted from data/welcome.md or data/welcome.html (if exists)
* display starred spaces

In coretiles, add a *star* icon/action to spaces to add them to user welcome page


osallou:
* user can access as visitor to public spaces
* user can see modules set to visitor
* user can search for spaces within space/catalog/resource name/descriptio
Screenshot from 2021-11-09 12-23-19
ns

PR #386, opened 2021-11-08
improve users form controls

gaetanHerve:


Modules/core/Controller/CoreusersController.php
```diff
@@ -125,33 +125,35 @@ public function editAction($id) {
         $script = "";
         if ($form->check()) {
             $canEditUser = true;
+
+            if(!$form->getParameter("email") || !$modelUser->isEmailFormat($form->getParameter("email"))) {
+                $canEditUser = false;
+                $_SESSION["flash"] = CoreTranslator::EmailInvalid($lang);
+                $_SESSION["flashClass"] = "danger";
```
osallou, line 132: flash message will be overriden by next controls, should "stop" here

osallou (changes requested):

osallou (approved):

PR #405, opened 2021-11-19
add menu space access controls

osallou:
checks if user has access to related menu


osallou:
what is missing ?

Many edit/delete actions do not check if user is owner of the "resource" (booking, quote, ...)
for space manager access only this is not a pb, for where user has access, ALL those action should check if user is owner of resource or if user is space manager to allow it

PR #408, opened 2021-11-23
add PfmUserException class

### LINKED ISSUE #406: [PfmExceptions] Create a PfmUserException class
Which would allow to filter exceptions linked to user as "this login already exists" and avoid it to be sent to Sentry.

gaetanHerve:
closes #406

### DIFF
--- a/Framework/Router.php
+++ b/Framework/Router.php
@@ -364,13 +364,8 @@ class Router {
      *        	Thrown exception
      */
     private function manageError(Throwable $exception, $type = '') {
-        $sendToSentry = true;
 
-        if ($exception instanceof PfmAuthException) {
-            $sendToSentry = false;
-        }
-
-        if($sendToSentry && Configuration::get('sentry_dsn', '')) {
+        if($exception->sendReports && Configuration::get('sentry_dsn', '')) {
             \Sentry\captureException($exception);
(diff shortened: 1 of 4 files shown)


Framework/Router.php
```diff
@@ -364,13 +364,8 @@ private function createController($urlInfo, Request $request) {
...
      */
     private function manageError(Throwable $exception, $type = '') {
-        $sendToSentry = true;
 
-        if ($exception instanceof PfmAuthException) {
-            $sendToSentry = false;
-        }
-
-        if($sendToSentry && Configuration::get('sentry_dsn', '')) {
+        if($exception->sendReports && Configuration::get('sentry_dsn', '')) {
```
osallou, line 368: nope, depending on exception, sendReports will not exists (php, db, etc. exception)

gaetanHerve: sendReports exists by inheritance (as property of PfmException) and is overridden when specified in child class. No ?

osallou: for pfmexceptions yes, not for other exceptions

gaetanHerve: ah, yeah of course

osallou:
There is merge conflict

PR #427, opened 2021-12-02
change menu to structure in coreTranslator

gaetanHerve:
I replaced "menu" by "structure" in strings generated by coreTranslator functions, but didn't renames the functions themselves (would imply to rename all references). But I could do that

I also replaced "menu" by "module" for module configurations. That was ambiguous.


osallou:
personally I would have changed only the "translation", not the function names etc...
If we decide to change *structure* to something else won't be relevant anymore and will need to change all again....   And , from code point of view this is indeed a menu....

Modules/core/Model/CoreInstall.php
```diff
@@ -642,6 +642,11 @@ public function upgrade_v3_v4() {
                 $eventHandler->customerImport();
                 Configuration::getLogger()->debug('[stats] import clients, done!');
             }
+
+            Configuration::getLogger()->debug("[core_adminmenu] change menu to structure");
+            $sql = "UPDATE core_adminmenu SET name='Structures' WHERE name='Menus'";
```
osallou, line 647: why changing in db ?

osallou: there are in coreadminemnus some checks about this , only translator should be modified

gaetanHerve: to have "structure" instead of "menu" in admin menu too:

admin_menu

Future superadmins maybe won't know that "structure" was "menu" at first...

PR #428, opened 2021-12-02
Feature plans

osallou:
introduce space plans with feature flags according to plan

plan cannot yet be modified via UI, all are in free plan...

additional plan config done via conf.yaml file in Config (see conf.yaml.sample)


osallou:
screenshot of plan page

plan_page

PR #440, opened 2021-12-09
Avoid missing files opening

### LINKED ISSUE #344: [Documents] Should check if file exists before trying to open it
cf [Sentry] (https://sentry.genouest.org/sentry/pfm/issues/428/)
Trying to simply raise an exception won't work because of open button onclick function:
```
<button type="button" onclick="location.href='documentsopen/23//21'" class="btn btn-xs btn-default">Open</button>
```
Clicking on "open" will load another page, avoiding to display the exception page.

gaetanHerve:
closes #344


Modules/documents/Controller/DocumentslistController.php
```diff
@@ -101,16 +101,19 @@ public function openAction($id_space, $id) {
...
-        header("Content-Description: File Transfer");
-        header('Content-Disposition: attachment; filename="'.basename($file).'"' );
-        header("Content-Type: application/zip");
-        header("Content-Transfer-Encoding: binary");
-
-        // read the file from disk
-        readfile($file);
-        //    return;
-        //$this->redirect($path);
+        if ($file && $file != "") {
```
osallou, line 104: If no file, what will occur?
No exception nor view i think

Why not just a if no file raise exception

gaetanHerve: it's more user friendly to have a flash message instead of an exception display... I'm gonna replace the condition with a file_exist($file) and add a try / catch raising an exception

osallou: flash is fine for me

osallou: I missed the else condition

gaetanHerve: ok... I just change the condition though and push back

Modules/documents/Controller/DocumentslistController.php
```diff
@@ -95,22 +95,23 @@ public function editAction($id_space, $id) {
...
-
-        // read the file from disk
-        readfile($file);
-        //    return;
-        //$this->redirect($path);
+        if (file_exists($file)) {
+            header("Cache-Control: public");
+            header("Content-Description: File Transfer");
+            header('Content-Disposition: attachment; filename="'.basename($file).'"' );
+            header("Content-Type: application/zip");
```
osallou, line 106: we are not sending app/zip here.... should be  binary/octet-stream

PR #441, opened 2021-12-10
Feature formrefactor

osallou:
refactor forms to remove large number of parameters and cleanup bootstrap usage


osallou:
not fully compliant, so should be used for new forms or update one after the other...

osallou:
makes use of externals/pfm/controls js library

PR #442, opened 2021-12-10
Feature refactor space mngt

osallou:
refactor controllers to load space at controller setup using space_id from input route if present

* get space object available directly rather than X calls
* avoid some duplications


osallou:
Adds to controller:

* currentSpace: space object if id_space is defined
* role: user role in space

Adds to views:

* context: with mainmenu,sidemenu,spacemenu,currentspace and role objects

PR #446, opened 2021-12-10
fix documents duplication on edition

### LINKED ISSUE #445: [Documents] Editing a document will add a new one
Editing a document will add a new one instead of modifying the existing one

gaetanHerve:
closes #445


Modules/documents/Controller/DocumentslistController.php
```diff
@@ -63,28 +63,26 @@ public function indexAction($id_space) {
...
         $form->addText("title", DocumentsTranslator::Title($lang), true, $data["title"]);
         $form->addUpload("file_url", DocumentsTranslator::File($lang));
-        $form->setValidationButton(CoreTranslator::Save($lang), "documentsedit/" . $id_space);
+        $form->setValidationButton(CoreTranslator::Save($lang), "documentsedit/" . $id_space . "/" . $id);
         $form->setCancelButton(CoreTranslator::Cancel($lang), "documents/" . $id_space);
 
         if ($form->check()) {
             $title = $this->request->getParameter("title");
             $id_user = $_SESSION["id_user"];
             $idNew = $model->set($id, $id_space, $title, $id_user);
```
osallou, line 79: don't you create a new entry in db here?

Modules/documents/Controller/DocumentslistController.php
```diff
@@ -63,28 +63,26 @@ public function indexAction($id_space) {
...
-
         $model = new Document();
         $data = $model->get($id_space, $id);
 
         $form = new Form($this->request, "DocumentsEditAction");
         $form->setTitle(DocumentsTranslator::Edit_Document($lang));
         $form->addText("title", DocumentsTranslator::Title($lang), true, $data["title"]);
         $form->addUpload("file_url", DocumentsTranslator::File($lang));
-        $form->setValidationButton(CoreTranslator::Save($lang), "documentsedit/" . $id_space);
+        $form->setValidationButton(CoreTranslator::Save($lang), "documentsedit/" . $id_space . "/" . $id);
```
osallou, line 73: why ? once edited we should go back to listing

gaetanHerve: no, we pass through editAction a second time, going into if ($form->check()) block, then are being redirected to list with $this->redirect("documents/" . $id_space) line 88. But now we have the document id (with this line, 73) and can tell if it's a new document or an existing one

Modules/documents/Controller/DocumentslistController.php
gaetanHerve: (in  if ($form->check()) block:
if $id > 0 => update existing entry entry
else => create new entry

osallou: so idnew will be the new id or the existing id, no ? at least this is what set method should do

osallou: and in this case, no need of iddoc

gaetanHerve: ah yes, I created $idDoc before passing $id in the request parameters... But I can remove it

PR #451, opened 2021-12-13
add structures names to displayed spacenames in account creation

### LINKED ISSUE #450: [BUG] demande d'inscription à l'espace ressources communes de Biosit
Bonjour,

Nous avons depuis quelques jours de nombreuses demandes d'accès à l'espace "ressources communes". Cette espace est dans le menu Biosit mais on a des demandes qui semblent venir de personnes hors biosit. Je me pose la question de l'affichage de cet espace dans un autre menu.
Stéphanie

gaetanHerve:
closes #450

### DIFF
--- a/Modules/core/Controller/CoreaccountController.php
+++ b/Modules/core/Controller/CoreaccountController.php
@@ -6,6 +6,9 @@ require_once 'Framework/Configuration.php';
 
 require_once 'Modules/core/Model/CoreConfig.php';
 require_once 'Modules/core/Model/CoreSpace.php';
+require_once 'Modules/core/Model/CoreMainMenu.php';
+require_once 'Modules/core/Model/CoreMainSubMenu.php';
+require_once 'Modules/core/Model/CoreMainMenuItem.php';
 require_once 'Modules/core/Model/CoreUser.php';
 require_once 'Modules/core/Model/CorePendingAccount.php';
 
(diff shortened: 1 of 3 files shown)


Modules/core/Controller/CoreaccountController.php
```diff
@@ -111,6 +111,14 @@ public function indexAction() {
         $modelSpaces = new CoreSpace();
         $spaces = $modelSpaces->getForList();
 
+        // add structures names to spaces displayed names
+        for($i=0; $i<count($spaces['ids']); $i++) {
```
osallou, line 115: not efficient: select from select from select and in a loop....
why not getting all menus and matching space id with structure name

gaetanHerve: yes, that's why I thought about just after having pushed. There are obviously too many sql request here. I'm gonna fix that

Modules/core/Controller/CoreaccountController.php
```diff
@@ -97,6 +100,47 @@ public function waitingAction() {
...
+    protected function addMenuNamesToSpaceNames($spacesList) {
+        $modelMenus = new CoreMainMenu();
+        $modelSubMenus = new CoreMainSubMenu();
+        $modelMenuItems = new CoreMainMenuItem();
+
+        $menus = $modelMenus->getAll();
+        $subMenus = $modelSubMenus->getAll();
+        $items = $modelMenuItems->getAll();
+        
+        for($i=0; $i<count($spacesList['ids']); $i++) {
```
osallou, line 117: if no spacelist yet , will fail

PR #453, opened 2021-12-15
Remove submit on change from user select in reservation page

gaetanHerve:


gaetanHerve:
maybe we should wait... It seems we have an unexpected behaviour with this modification... I have to make some tests

gaetanHerve:
OK. submit on change active on user select was meant to avoid reloading the whole page for just uploading clients list when changing user. I'm gonna use my dynamic forms external here

gaetanHerve:
OK now

Modules/booking/View/Bookingdefault/addreservationAction.php
```diff
@@ -7,6 +7,19 @@
...
     <?php echo $form->htmlOpen() ?>
     <?php echo $form->getHtml($lang, false) ?>
+    <script type="module">
+        import {DynamicForms} from '/externals/pfm/dynamics/dynamicForms.js';
+        let dynamicForms = new DynamicForms();
+        let spaceId = <?php echo $id_space?>;
+        let sourceId = "recipient_id";
+        let targets = [
+            {
+                elementId: "responsible_id",
```
osallou, line 17: Source and target identical?

osallou:
And if user (and not manager), user also select client?

Modules/booking/View/Bookingdefault/addreservationAction.php
gaetanHerve: euh... "recipient" and "responsible" are not identical :)

gaetanHerve:
> And if user (and not manager), user also select client?

Yes, but only amongst the clients they are related to. If there are several clients for a user, a choice must be done somewhere...
By default, the first of user's clients is selected.

Modules/booking/View/Bookingdefault/addreservationAction.php
osallou: Bad reading.... i should rest

PR #463, opened 2021-12-17
clean coreconfigadmin menus

gaetanHerve:


Modules/core/Controller/CoresecureController.php
```diff
@@ -79,7 +79,7 @@ protected function checkRememberMeCookie() {
      */
     public function runAction($module, $action, $args = array()) {
         $modelConfig = new CoreConfig();
-        if ($modelConfig->getParam("is_maintenance") && ($this->request->getSession()->getAttribut("user_status") < 4)) {
+        if ($modelConfig->getParam("is_maintenance") && ($this->request->getSession()->getAttribut("user_status") < 2)) {
```
osallou, line 82: rather than 2, 4, should use existing variables (easier to read understand) such as CoreStatus::$ADMIN

osallou: but there are different values among CoreSpace, CoreStatus and CoreUser.
CoreSpace is for user roles in space, but CoreStatus:$ADMIN vs CoreUser:$ADMIN ???

Modules/core/Controller/CoresecureController.php
```diff
@@ -79,7 +79,7 @@ protected function checkRememberMeCookie() {
      */
     public function runAction($module, $action, $args = array()) {
         $modelConfig = new CoreConfig();
-        if ($modelConfig->getParam("is_maintenance") && ($this->request->getSession()->getAttribut("user_status") < 4)) {
+        if ($modelConfig->getParam("is_maintenance") && ($this->request->getSession()->getAttribut("user_status") < CoreStatus::$ADMIN)) {
                 throw new PfmException($modelConfig->getParam("maintenance_message"), 503);
```
osallou, line 83: should raise an exception not sent to sentry (PfmUserException ?)

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changes
 
+## 2.1.1
```
osallou, line 3: will be 2.1.2 as we already delivered some 2.1.1

PR #475, opened 2021-12-21
Clarify selfregistration messages

gaetanHerve:
- wrong login or password was silently failing
- changing account validation process was invalidating message sent to users

### DIFF
--- a/Modules/core/Controller/CoreconnectionController.php
+++ b/Modules/core/Controller/CoreconnectionController.php
@@ -99,9 +99,6 @@ class CoreconnectionController extends CorecookiesecureController {
      */
     public function loginAction() {
         $lang = $this->getLanguage();
-        $modelConfig = new CoreConfig();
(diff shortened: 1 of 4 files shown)


Modules/core/Controller/CoreconnectionController.php
```diff
@@ -110,49 +108,63 @@ public function loginAction() {
...
+                $msg = CoreTranslator::AccountInactive($lang);
+                break;
+            case "invalid_password":
+                $msg = CoreTranslator::InvalidPassword($lang);
+                break;
+            case "invalid_login":
+                $msg = CoreTranslator::InvalidLogin($lang);
+                break;
+            default:
+                $msg = $connection_error;
```
osallou, line 161: default is empty.... should at least display a "something went wrong" or alike

gaetanHerve: ah yes. At first $connection_error parameter was not optional. I add that

PR #478, opened 2021-12-21
Fix booking role related auth

gaetanHerve:


Modules/booking/Controller/BookingdefaultController.php
```diff
@@ -129,13 +129,22 @@ private function canUserEditReservation($id_space, $id_resource, $id_user, $id_r
...
-        $modelSpace = new CoreSpace();
-        $role = $modelSpace->getUserSpaceRole($id_space, $_SESSION["id_user"]);
-        if (!$isUserAuthorizedToBook && $role < CoreSpace::$MANAGER) {
+        $modelBkAccess = new BkAccess();
+        $bkAccess = $modelBkAccess->getAccessId($id_space, $resource['id']);
+
+        $curentDate = $_SESSION['bk_curentDate'];
+        $curentDate = (!$curentDate || $curentDate == "") ? date("Y-m-d") : $curentDate;
+        $temp = explode("-", $curentDate);
+        $curentDateUnix = mktime(0, 0, 0, $temp[1], $temp[2], $temp[0]);
```
osallou, line 143: should check bk_curentDate may not be in session

 $curentDate = date("Y-m-d", time());
        if (isset($_SESSION['bk_curentDate'])) {
           $curentDate = $_SESSION['bk_curentDate'];
        }

even if I am not fond of this date in session, and should be anough to check against "today"

PR #482, opened 2022-01-03
Fix mktime 481

osallou:


gaetanHerve (approved):
Seems ok to me. Made manual tests to check interface behaviour on booking module and to get the new exceptions.

PR #501, opened 2022-01-07
Fix performance

osallou:
fix perf, reducing number of sql request from ~200 to ~25....
affects booking dayarea et weekarea


gaetanHerve (approved):
Ok to me

PR #506, opened 2022-01-11
Fix services edition units

### LINKED ISSUE #412: [Services edition] Bug if user sets non-numeric values in quantity/time/price input
Generating global statistics, I had an error coming from a"2h -" value set in quantity/time/price input (service['quantity']).
(SeProject.php on line 788)
We should:
- forbid user to input non-numeric values
- clarify type of unit management for services. "quantity/time/price" is kind of unclear and motivates the add of non-numeric details

gaetanHerve:
closes #412


Framework/FormHtml.php
```diff
@@ -460,9 +459,8 @@ static public function inlineLabel($name, $value, $vect = false) {
         if ($vect) {
             $vectv = "[]";
         }
-        $html = "<span class=\"form-control\" id=\"" . $name . "\" name=\"" . $name . $vectv . "\"";
-        $html .= ">" . $value . "</span>";
-
+        $html = "<label class=\"label\" style=\"color:black;\" id=\"" . $name . "\" name=\"" . $name . $vectv . "\"";
```
osallou, line 462: why color black?

osallou: but others do not have any style

PR #521, opened 2022-01-14
Feature userspaces

osallou:
Add a button with structure "my spaces" , will display all spaces a user is member of

### DIFF
--- a/Modules/core/Controller/CoretilesController.php
+++ b/Modules/core/Controller/CoretilesController.php
@@ -130,12 +130,21 @@ class CoretilesController extends CorecookiesecureController {
 
             $allSpaces = $spaceModel->getSpaces('id');
             $spaceMap = [];
+            $userSpacesOnly = $this->request->getParameterNoException('mine') === '1' ? true: false;
+            if($userSpacesOnly) {
+                $spaces= [];
+            }
+            $userSpacesList = [];
+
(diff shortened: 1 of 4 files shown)


Modules/core/Controller/CoretilesController.php
```diff
@@ -130,16 +130,27 @@ public function showMainSubMenu($id){
...
                             $space['join'] = CoreTranslator::RequestJoin($isMemberOfSpace, $lang);
                         }
+                        if($userSpacesOnly) {
+                            $userSpacesList[] = $space;
+                        }
                     } else {
                         $space['join_requested'] = CoreTranslator::JoinRequested($lang);
                     }
+                } else if($userSpacesOnly && $logged && in_array($space["id"], $userSpaces['spacesUserIsAdminOf']) && isset($_SESSION["login"])) {
+                        $userSpacesList[] = $space;
```
gaetanHerve, line 153: didn't you want to make an array_push($userSpacesList, $space) here ?

gaetanHerve: ah ok. condition should be if($userSpacesOnly && $isMemberOfSpace) I think

gaetanHerve (changes requested):
see inline comments. Tests returned all spaces

PR #528, opened 2022-01-18
use dynamicForms in project edition

gaetanHerve:


Modules/services/Controller/ServicesprojectsController.php
```diff
@@ -537,14 +537,14 @@ public function editAction($id_space, $id) {
...
+        $clients = $modelClient->getForList($id_space);
 
         $modelVisa = new SeVisa();
         $inChargeList = $modelVisa->getForList($id_space);
 
-        //$form->addSeparator(CoreTranslator::Description($lang));
         $form->addSelectMandatory("in_charge", ServicesTranslator::InCharge($lang), $inChargeList["names"], $inChargeList["ids"], $value["in_charge"]);
-        $form->addSelectMandatory("id_resp", CoreTranslator::Responsible($lang), $resps["names"], $resps["ids"], $value["id_resp"]);
+        // id_client is denominated id_resp in se_project table
+        $form->addSelectMandatory("id_client", ClientsTranslator::ClientAccount($lang), $clients["names"], $clients["ids"], $value["id_resp"]);
```
osallou, line 547: impacts unit tests to change the name of the parameter, so need to fix tests too...

PR #529, opened 2022-01-18
add script to setup a demo site with preconfigured data

osallou:


osallou:
TODO: force manager and demo passwords (so that user can log with those)

osallou:
Need to add spam filter on user entries

PR #532, opened 2022-01-19
[booking] send email only if user status > VISITOR Closes #531

### LINKED ISSUE #531: [Booking] Cancellation emails still being sent to inactive user
A former user notified me she was still receiving emails for cancelled slots on the MRic Photonics facility (Ressource: Confocal SP8 in her case) even after I switched her status to "Inactive".
Is this normal behaviour and if not is there an easy fix?
Thanks!
Xavier

osallou:


gaetanHerve (approved):
Tested and OK

Issue #290, opened 2021-09-22
[Login] Use of Renater Identity Federation


gaetanHerve:
We already discussed this subject, but there's no track of it. This question seems to be significant to several users.
@osallou, you told me about some difficulties in setting this up. Could you remind it here ?

osallou:
No openid support, need to setup saml (shiboleth) support
It is more complex to setup on apache proxy, and need for specific code to handle remote user request header.

And getting renater auth access requires quite long administrative tasks

So can be done, but longer for prod setup, specific proxy setup etc....

PR #533, opened 2022-01-20
[core][auth] add shibboleth support

osallou:
see doc in docker/README.md

manage user authentication but can also be used to create an account


osallou:
Need to declare service and test in prod

osallou:
linked to #290

PR #535, opened 2022-01-21
Record and show mailer messages

osallou:
Mailer can now be enabled at user level
Superadmins can send email to all spaces admins

* if user will only display messages sent by space managers to space users
* if user, mail submission is hidden and disabled
* if manager, will see all messages
* if superadmin, can send email to admins of all spaces

This gives an history of mails and user can view previous sent
messages


osallou:
this would also enable us (superadmin) in a dedicated space to send mails to other spaces admins

Modules/mailer/Controller/MailerController.php
```diff
@@ -56,21 +58,54 @@ public function indexAction($id_space) {
...
     }
 
     public function sendAction($id_space) {
         $this->checkAuthorizationMenuSpace("mailer", $id_space, $_SESSION["id_user"]);
+
+        $modelConfig = new CoreConfig();
+        $editRole = $modelConfig->getParamSpace("mailerEdit", $id_space, CoreSpace::$ADMIN);
+        
+        if($this->role < $editRole) {
+            throw new PfmAuthException('not enough priviledges');
```
gaetanHerve, line 100: typo: "privileges"

gaetanHerve: Used in feature_doc_visibility too (and I suppose in other PRs in the future). Maybe we can factorize that into a single shared method... On the other hand, it's only 2 lines...

osallou: for single line test does not worth calling an other method, just checking role

PR #538, opened 2022-01-24
create unique vue for user space booking and vclient options

gaetanHerve:


osallou:
once done (or almost) please add some screenshot to see hat it looks like

PR #547, opened 2022-02-01
[invoices] Feature async

osallou:
* manage invoices in background (to avoid long calculations delay)
*factorize invoice computations, were in 2 places in xxxinvoice controllers and xxinvoice in model depending on where request was done (global or specific)
* fixes on checks (global were checking only for booking clients)
* incoherent code between different invoices implementations


Framework/Events.php
```diff
@@ -24,10 +25,16 @@
 require_once 'Modules/quote/Model/Quote.php';
 require_once 'Modules/services/Model/SeService.php';
 
+require_once 'Modules/invoices/Model/GlobalInvoice.php';
+
+require_once 'Modules/booking/Model/BookingInvoice.php';
+require_once 'Modules/services/Model/ServicesInvoice.php';
+
```
gaetanHerve, line 32: missing import: require_once 'Modules/booking/Model/BookinginvoiceTranslator.php';

osallou: not needed here, certainly elsewhere, i gonna look

osallou: ah confusing with translators.... ok

osallou: in bookinginvoice.php

Modules/booking/Model/BookingInvoice.php
```diff
@@ -281,11 +465,11 @@ protected function getUnitPackagePricesForEachResource($id_space, $resources, $L
         $modelPriceOwner = new BkOwnerPrice();
         foreach ($resources as $resource) {
             // get the packages prices
-            $packages = $modelPackage->getByResource($id_space, $resource["id"]);
+            $packages = $modelPackage->getByResource($id_space ,$resource["id"]);
 
             $pricesPackages = array();
             for ($i = 0; $i < count($packages); $i++) {
-                $price = $modelPriceOwner->getPackagePrice($id_space, $packages[$i]["id"], $resource["id"], $id_unit);
+                $price = $modelPriceOwner->getPackagePrice($id_space, $packages[$i]["id"], $resource["id"], $id_client);
```
osallou, line 472: To check this modification, why changing unit to client, is it an error?

osallou (commented):
To check

Modules/booking/Model/BookingInvoice.php
osallou: nope, there is no unit defined

gaetanHerve:
closes #657

PR #551, opened 2022-02-02
Feature async stats

osallou:
create stats in background


gaetanHerve:
should add a refresh button, like  in async invoices

osallou:
(attention au restart du pfm-events)

gaetanHerve (approved):
all seems to work well

osallou:
issues are not important, but I pushed some cleanup

PR #560, opened 2022-02-08
Feature cal refactor

osallou:
internal rewriting of booking calendars

* fixes small screens layout keeping real horizontal layout
* fixes WAI ARIA related stuff (set cells columns relations, some labels etc...)
* refactoring: display menu bar and compute/generate visual calendar in a single place instead of duplicated code (90% equals) in 5 different places (views for day, dayarea etc...)


osallou:
also integrates #558 updates, deprecating it

PR #562, opened 2022-02-09
Feature bksched scale

osallou:
Handle booking settings for resa_time_setting and booking_time_scale
    
Those were defined but not use.
    
Impacts scheduling add/edit to show end date or duration, and sets default end time defaults


gaetanHerve (approved):
tests ok

PR #564, opened 2022-02-09
[booking] add to scheduling option to force usage of packages to an area

osallou:
relates to #563

in bookingscheduling, add option force_packages

if set, on new reservations, user must select a package


osallou:
Maybe this should be defined at area level in booking settings

One may want to force package only for some resources but not all

gaetanHerve (approved):
ok to me. 
left changelog update and conflicts resolution

Issue #549, opened 2022-02-02
[core] When deactivating a module, module options are still accessible in space users menu


gaetanHerve:
- Deactivating a module should set related core_space_access_options to deleted = 1
- Reactivating a module should set related existing core_space_access_options to deleted = 0

PR #568, opened 2022-02-15
Fix module deactivation

gaetanHerve:
Once activated, setting a module to "inactive" wouldn't remove its related menu to users' space access options. (see #549 )
Fixed with this PR, setting deleted=1 for its core_space_access_options entry at deactivation.

### DIFF
--- a/Modules/booking/Controller/BookingconfigController.php
+++ b/Modules/booking/Controller/BookingconfigController.php
@@ -55,7 +55,7 @@ class BookingconfigController extends CoresecureController {
                 $modelAccess->exists($id_space, $toolname)
                     ? $modelAccess->reactivate($id_space, $toolname)
                     : $modelAccess->set($id_space, $toolname, "booking", $toolname);
-            } else if ($modelAccess->exists($id_space, $toolname)) {
+            } else {
                 $modelAccess->delete($id_space, $toolname);
             }
             return $this->redirect("bookingconfig/".$id_space);
--- a/Modules/clients/Controller/ClientsconfigController.php
+++ b/Modules/clients/Controller/ClientsconfigController.php
@@ -51,7 +51,7 @@ class ClientsconfigController extends CoresecureController {
                 $modelAccess->exists($id_space, $toolname)
                     ? $modelAccess->reactivate($id_space, $toolname)
                     : $modelAccess->set($id_space, $toolname, "clients", $toolname);
-            } else if ($modelAccess->exists($id_space, $toolname)) {
+            } else {
                 $modelAccess->delete($id_space, $toolname);
             }
 
--- a/Modules/core/Model/CoreSpaceAccessOptions.php
+++ b/Modules/core/Model/CoreSpaceAccessOptions.php
@@ -38,7 +38,7 @@ class CoreSpaceAccessOptions extends Model {
             $this->runRequest($sql, array($id_space, $toolname, $module, $url));
         }
         else{
-            $sql = "UPDATE core_space_access_options SET module=?, url=?, WHERE id_space=? AND toolname=?";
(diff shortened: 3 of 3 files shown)


Modules/core/Model/CoreSpaceAccessOptions.php
```diff
@@ -28,18 +28,18 @@ public function __construct() {
     }
     
     public function getAll($id_space){
-        $sql = "SELECT * FROM core_space_access_options WHERE id_space=?";
+        $sql = "SELECT * FROM core_space_access_options WHERE id_space=? AND deleted=0";
         return $this->runRequest($sql, array($id_space))->fetchAll();
     }
     
-    public function set($id_space, $toolname, $module, $url){
+    public function set($id_space, $toolname, $module, $url, $deactivate=0) {
```
osallou, line 35: why using delete param on set  and not calling delete (or a delete like based on toolname/module)

gaetanHerve: ah yes, would be more coherent

gaetanHerve: done

Modules/core/Model/CoreSpaceAccessOptions.php
```diff
@@ -28,17 +28,17 @@ public function __construct() {
...
     
-    public function set($id_space, $toolname, $module, $url){
+    public function set($id_space, $toolname, $module, $url) {
         if (!$this->exists($id_space, $toolname)){
             $sql = "INSERT INTO core_space_access_options (id_space, toolname, module, url) VALUES (?,?,?,?)";
             $this->runRequest($sql, array($id_space, $toolname, $module, $url));
         }
         else{
-            $sql = "UPDATE core_space_access_options SET module=?, url=? WHERE id_space=? AND toolname=?";
+            $sql = "UPDATE core_space_access_options SET module=?, url=?, WHERE id_space=? AND toolname=?";
```
osallou, line 41: No comma here

Modules/clients/Controller/ClientsconfigController.php
```diff
@@ -46,8 +46,15 @@ public function indexAction($id_space) {
...
             $this->menusactivation($id_space, 'clients', 'credit-card');
             $modelAccess = new CoreSpaceAccessOptions();
-            $modelAccess->set($id_space, "clientsuseraccounts", "clients", "clientsuseraccounts");
-                
+            $toolname = "clientsuseraccounts";
+            if ( $this->request->getParameter("clientsMenustatus") > 0 ) {
+                $modelAccess->exists($id_space, $toolname)
+                    ? $modelAccess->reactivate($id_space, $toolname)
+                    : $modelAccess->set($id_space, $toolname, "clients", $toolname);
+            } else if ($modelAccess->exists($id_space, $toolname)) {
```
osallou, line 54: If not really neeeed. Will try delete and won t do anything if not exists

Modules/booking/Controller/BookingconfigController.php
```diff
@@ -40,21 +40,24 @@ public function indexAction($id_space) {
...
-            }
+            $this->menusactivation($id_space, 'bookingsettings', 'calendar', 'booking');                   
             
+            $modelAccess = new CoreSpaceAccessOptions();
+            $toolname = "bookingauthorisations";
+            if ( $this->request->getParameter("bookingsettingsMenustatus") > 0 ) {
+                $modelAccess->exists($id_space, $toolname)
+                    ? $modelAccess->reactivate($id_space, $toolname)
+                    : $modelAccess->set($id_space, $toolname, "booking", $toolname);
+            } else if ($modelAccess->exists($id_space, $toolname)) {
```
osallou, line 58: Same as below comment if not needed

PR #578, opened 2022-02-25
[booking] manage shared area calendar option to conflict resources on same calendar

osallou:
see #404 

* option on booking calendar to share calendar among resources in a same area
* if enabled, resources booking will conflict in a same area
* booking calendar display will prevent booking if already booked by an other resource in same area


gaetanHerve (approved):
Tested before and after merging develop. OK to me

PR #589, opened 2022-03-02
Fix tile w/o space display

gaetanHerve:


Modules/core/Controller/CoretilesController.php
```diff
@@ -207,6 +207,14 @@ public function showMainSubMenu($id){
             $title = $modelSubMenu->getName($id);
         }
 
+        // filter out items with no space
+        $refLength = count($items);
+        for ($i=0; $i<$refLength; $i++) {
+            if (!$items[$i]) {
+                unset($items[$i]);
```
osallou, line 214: if you unset, next unset will be at wrong index at items has been modified meanwhile

gaetanHerve: ah yes, doesn't handle the case where there's 2 items to unset in a row

gaetanHerve:
I should also handle items deletion at ACTION_SPACE_DELETE event to complete this PR

osallou:
> I should also handle items deletion at ACTION_SPACE_DELETE event to complete this PR

why? what do you want to do ?

Modules/core/Controller/CoretilesController.php
gaetanHerve: Because if a space is deleted, we don't need its tile anymore, unless we remember it exists and use it for the next space we create

gaetanHerve:
Because if a space is deleted, we don't need its tile anymore, unless we remember it exists and use it for the next space we create

osallou:
ready for merge?

gaetanHerve:
yes, if you had a check at my last comment

PR #593, opened 2022-03-04
use bcrypt algo for passwords

osallou:
keep compat with existing ones, setting… algo new core_users column to md5

add passwordEncode and comparePasswords methods in core_users, with hash parameter.
Default is to use bcrypt for new/modifications.

use jquery npm instead of copied unversioned lib


osallou:
also remove old libs not used anymore

osallou:
to be tested with prod backup

gaetanHerve (commented):
why all that bootstrap deletions ?

osallou:
for bootstrap, not related, but use npm installed library  instead of copied local version

gaetanHerve (approved):
tried to connect with an existing user, then created one and been able to connect PFM. Seems ok to me

PR #597, opened 2022-03-07
[booking] add to calendar a summary view option

osallou:
Related to #591 comments

Read only view showing only bookings per day/resources without slots details.


osallou:
Screenshot from 2022-03-07 14-32-01

PR #604, opened 2022-03-14
[booking] list blocked resources and add a reason column on bkcalentry

osallou:
* In blocked resources, list blocked resources
* In bkcalentry add a column to store entry reason: booking, holiday, maintenance (booking is default)
* In calendar, if reason > 0 (not booking) display reason and short desc

This will show in calendar explicit holiday/maintenance... blocking

With list, managers have a view on all blocked resources rather than additional booking


Modules/booking/Model/BkCalendarEntry.php
```diff
@@ -904,6 +910,12 @@ public function journal(int $id_space, int $id_user, int $max=100) {
         return $res->fetchAll();
     }
 
+    public function blockedEntries($id_space) {
+        $sql = "SELECT * FROM bk_calendar_entry WHERE reason>0 AND id_space=? AND deleted=0 ORDER BY start_time DESC";
+        $req = $this->runRequest($sql, array($id_space));
+        return $req->fetchAll();
+    }
+
     public function computeDuration($id_space, $booking) {
```
gaetanHerve, line 919: clicking on edit on a blocked resource line in bookingblock/{id_space} will throw the PfmException('no pricing found for client '.$id_client) line 935. We could use a $booking['reason'] != 0 condition to handle this case then access to the reservationedit page

gaetanHerve (changes requested):
works well aside comment.

osallou:
@gaetanHerve j'a pushé un fix, conflit lié à l'affichage du détail de la résa
si pas de client, il affiche rien

gaetanHerve (approved):

PR #606, opened 2022-03-14
[booking] try to fix booking packages

osallou:


osallou:
* price and owner price stores package *id_package*
* bkcalentry stores package *id*
* invoice_item stores package *id*

PR #609, opened 2022-03-15
hide space unsubscribe button if user is pending or not member of space

gaetanHerve:


Modules/core/View/Corespace/viewAction.php
```diff
@@ -119,7 +119,7 @@
             <?php
         }
         ?>
-        <?php if($_SESSION['id_user'] > 0 && $role<CoreSpace::$MANAGER && $role > 0) { ?>
+        <?php if($_SESSION['id_user'] > 0 && $role<CoreSpace::$MANAGER && $role > 0 && $isMemberOfSpace && !$isPending) { ?>
```
osallou, line 122: if is member, he is not pending, no? ismemberofspace not enough?

PR #613, opened 2022-03-16
can delete bookingpackages lines with not unique package_id

gaetanHerve:


gaetanHerve:
I'm gonna refactor Bookingsupsinfo, Bookingcalquantities and Bookingpackages controllers using a parent class. Most of the mechanics are the same

osallou:
Warning: in packages we could have only one element per paclage id, this may not be the same for otherd

gaetanHerve:
yes, I'll be careful to keep the right behaviour

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 2.5.0
 
+* [booking_quantities] avoid having two qtes with the same name for the same resource
```
osallou, line 5: will be 2.6.0 now

Modules/booking/Controller/BookingsupsabstractController.php
```diff
@@ -0,0 +1,186 @@
...
+        return false;
+    }
+
+    protected function hasInvoicingUnitsDuplicates($supResources, $supIsInvoicingUnit, $id_space, $lang) {
+        $result = false;
+        $invoicingUnitsResources = [];
+        foreach ($supResources as $index => $resource) {
+            if ($supIsInvoicingUnit[$index] == 1) {
+                if (in_array($resource, $invoicingUnitsResources)) {
+                    $_SESSION["flash"] = BookingTranslator::maxInvoicingUnits($lang);
```
osallou, line 159: is message adapted to package/supp/....?

gaetanHerve: $invoicable property is set to false for supplementaries and packages. So it never pass through hasInvoicingUnitsDuplicates() (and these lines of code). cf line 101:
```
if ($this->invoicable && $this->hasInvoicingUnitsDuplicates($supResources, $supIsInvoicingUnit, $id_space, $lang))
```

osallou:
changelog conflict, else, should it be merged?

Issue #608, opened 2022-03-15
[Resources] When deleting a resource, delete related sup infos, packages and quantities


gaetanHerve:
Currently, when a resource is deleted, its related info sups, packages and quantities still are listed but display another resource in place of the deleted one. Could generate unwanted behaviour if space admins click save without having manually deleted them.

osallou:
still appear if not deleted too indeed, but how can it display an other resource ? Request is always related to a resource id

In some cases, for packages, there is a package_id link, a package_id can be the same for multiple resources (based on package name), but later requests on invoice etc.. will request for a package+resource, so should not be found.

osallou:
there are also prices etc.. I think

osallou:
for packages, it is a select. Value is the id of the deleted resource. As not anymore in the select options, the select takes the first resource found. So ok (for now) in db, but not in display. Related resource elements should be deleted too.

osallou:
Resource related stuff

* bk_access
* bk_authorization
* bk_calquantities
* bk_calsupinfo
* bk_owner_prices
* bk_packages
* bk_prices
* bk_restrictions
* re_event
* re_resps

For resource category:

* should not delete category if has linked resources
* re_visas

gaetanHerve:
thanks for this exhaustive list. The select problem is what i meant by "display another resource in place of the deleted one"

osallou:
just like in #611 , need to trigger removal in other modules with no resource code pollution (resource module does not have to know about other modules)

osallou:
modules could add in db (at install) events they are interested on and on pfm-events could trigger some related module function

for ex on event ACTION_RESOURCE_DELETE , pfm-events could trigger some call on modules listening for this

this is background task however, will not prevent resource deletion for example if a category exists

PR #623, opened 2022-03-30
prevent pricing removal if a client is using it

osallou:


osallou:
TODO:

* on client removal, remove user relationship to client
* on resource delete, remove bkauthorizations

osallou:
see #608 for the many links

PR #624, opened 2022-03-30
add getDefault to bknightwe class

gaetanHerve:


Modules/booking/Model/BkNightWE.php
```diff
@@ -24,6 +24,21 @@ public function __construct() {
...
+        return array(
+            "id" => "",
+            "id_belonging" => 0,
+            "tarif_unique" => 1,
+            "tarif_night" => 0,
+            "night_start" => 19,
+            "night_end" => 8,
+            "id_category" => 0,
+            "tarif_we" => 0,
+            "choice_we" => "",
```
osallou, line 37: not the same as invoice defaults with saturday and sunday

Modules/booking/Model/BkNightWE.php
```diff
@@ -24,6 +24,21 @@ public function __construct() {
...
 
+    public function getDefault() {
+        return array(
+            "id" => "",
+            "id_belonging" => 0,
+            "tarif_unique" => 1,
+            "tarif_night" => 0,
+            "night_start" => 19,
+            "night_end" => 8,
+            "id_category" => 0,
```
osallou, line 35: category?

gaetanHerve: hein. Don't ùknow how my test went right

Modules/booking/Model/BkNightWE.php
gaetanHerve: ? when not set, seems to be handled as saturday + sunday....

osallou: in invoice, this is set

gaetanHerve: maybe get the areascheduling

PR #636, opened 2022-04-22
Feature caldav

osallou:
Add caldav feature to export user bookings calendar (all resources/areas) to his calendar client

Caldav url: https://.../caldav/*id_space*

User uses his credentials


osallou:
Fails with btsync, PROPFIND requests are find, calendar discovery too, but no REPORT request....though no error log

osallou:
getting report now and works with btsync,but get only one event in calendar, though multiple evens are returned

PR #639, opened 2022-05-06
Fix corespaceuseraccess

osallou:
if clients or booking are in inactive state, calls in editactions to client/bookings are made, but user is not able to access it and throw exception (authorization), even space admin.

Also, corespaceacccess generateSpaceAccessForm and validateSpaceAccessForm required access to clients module, though there is no reason for that (edit user info, not client related)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * [invoice][booking] manage case where total hours = 0
 * [core] add login button on welcome page
 * [booking] for multi slot bookings, do not display next slot content
+* [core] fix corespaceuseraccess when clients or booking modules are in inactive status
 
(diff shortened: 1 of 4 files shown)


gaetanHerve:
I meant on users module. Since users module is an "admin module", it would be this check: $this->checkSpaceAdmin($id_space, $_SESSION["id_user"]);
But maybe not necessary for the reason I mentionned before

osallou:
> I meant on users module. Since users module is an "admin module", it would be this check: $this->checkSpaceAdmin($id_space, $_SESSION["id_user"]); But maybe not necessary for the reason I mentionned before

done in calling functions, so is enough for me

PR #648, opened 2022-05-18
Upgrade vuejs

gaetanHerve:
I had to browserify vue3, didn't manage to make it work from node_modules.
Modified views using vuejs to make it work with vue3.


osallou:
if broserify is needed, then installing via npm is not the good solution.
You should instead in this case download vue3 "source" (from https://unpkg.com/vue@3 ?) and save it along curent vuejs code, then import it the same way.

osallou:
@gaetanHerve je pense qu'on peut clore cette PR

PR #662, opened 2022-06-16
Fix get invoices by id instead of number 

gaetanHerve:


osallou (changes requested):
Delete, done with number, should also be updated

osallou (approved):

PR #669, opened 2022-06-27
fix task done and visibility type for db requests

gaetanHerve:


osallou:
can assign only users, not managers

Modules/services/Controller/ServicesprojectsController.php
```diff
@@ -677,6 +685,34 @@ public function getTaskFileAction($id_space, $id_task) {
         $this->render(['data' => $file]);
     }
 
+    public function openFileAction($id_space, $id_project, $id_task) {
+        $this->checkAuthorizationMenuSpace("services", $id_space, $_SESSION["id_user"]);
+        $taskModel = new SeTask();
+        $task = $taskModel->getById($id_space, $id_task);
+
+        // If private task, check if user is the owner of the task
+        if ($task['private'] == 1 && $task['id_owner'] != $_SESSION["id_user"]) {
```
osallou, line 694: If user yes, else any manager should ve allowed

Modules/services/Controller/ServicesprojectsController.php
```diff
@@ -677,6 +685,34 @@ public function getTaskFileAction($id_space, $id_task) {
...
+            header('Content-Disposition: attachment; filename="'.basename($file).'"');
+            header('Expires: 0');
+            header('Cache-Control: must-revalidate');
+            header('Pragma: public');
+            header('Content-Length: ' . filesize($file));
+            readfile($file);
+        } else {
+            $_SESSION['flash'] = DocumentsTranslator::Missing_Document($this->getLanguage());
+            $_SESSION['flashClass'] = "warning";
+            $this->redirect("servicesprojects/kanban" . $id_space . "/" . $id_project);
```
osallou, line 712: Should add to redirect a 404 code in case of api call

Modules/services/Model/SeTask.php
```diff
@@ -160,7 +160,7 @@ public function getTaskServicesIds($id_space, $id_task) {
 
     public function setTaskService($id_space, $id_task, $id_service) {
         if ($this->isTaskService($id_space, $id_task, $id_service)) {
-            $sql = "UPDATE se_task_service (id_space, id_task, id_service) VALUES (?, ?, ?)";
+            $sql = "UPDATE se_task_service SET id_space=?, id_task=?, id_service=?;";
```
osallou, line 163: Why a  set vs previous update. This is ok, just wondering why

gaetanHerve: I think I had a problem testing with sql in strict mode. But don't see why there would have been a problem on this line...

Modules/services/Controller/ServicesprojectsController.php
gaetanHerve: You mean we're ok with this behaviour ?

osallou: No

If private and role is user, then you need to check owner (what you do) but if role is >= manager, then you should have access

gaetanHerve: we already have the `$this->checkAuthorizationMenuSpace("services", $id_space, $_SESSION["id_user"]);` above

osallou: If private and we are manager, id owner may be different from session id user.

A manager creates the task/file, an other manager wants to access it

gaetanHerve: I can modify the condition to "if role >= manager OR if session user is the task owner"

osallou: We won t allow users (members of project of course) to create task.
Only to view them, if public, and later possibly comment on task or change their status.

Only managers can create tasks

Users must not be allowed to create/edit tasks for the moment

gaetanHerve: I wanted to deal with authorising users to view projects in a future PR.
Currently, only roles authorised to access to services (generally managers and admins) can access to projects (read and write, no difference right now).

Modules/services/Controller/ServicesprojectsController.php
```diff
@@ -632,7 +641,8 @@ public function kanbanAction($id_space, $id_project) {
             "projectServices" => json_encode($services),
             "projectUsers" => json_encode($projectUsers),
             "mainUser" => $projectMainUser,
-            "personInCharge" => $project['in_charge']
+            "personInCharge" => $project['in_charge'],
+            "userIsManager" => json_encode($isManager),
```
osallou, line 645: why json encode a boolean?

Modules/services/Controller/ServicesprojectsController.php
```diff
@@ -686,6 +706,33 @@ public function getTaskFileAction($id_space, $id_task) {
...
     }
 
+    public function openFileAction($id_space, $id_task) {
+        $this->checkAuthorizationMenuSpace("services", $id_space, $_SESSION["id_user"]);
+        $taskModel = new SeTask();
+        $task = $taskModel->getById($id_space, $id_task);
+
+        // If private task, check if user is the owner of the task or if user is at least manager
+        if ($task['private'] == 1
+            && (!$this->role >= CoreSpace::$MANAGER || $task['id_owner'] != $_SESSION["id_user"])) {
```
osallou, line 716: invalid test I think   if not manager or not owner , the "not owner" will match  if i am a manager but not the owner

if .... && ( $this->role < CoreSpace::$MANAGER && $task['id_owner'] != $_SESSION["id_user"] )

no?

osallou (changes requested):
there is still a few pb
by the way, I understand that you can add only one file per task, this is limiting. You should be able to add multiple (and delete them...)

And to ease stuff, we should be able to list all documents of the project (rather than searching through all the tasks).

To go to prod for now, I suggest you remove file mngt and put them in a later development (even in schema as you add a file and file_name in task schema (which won't fit for multiple files).

Modules/services/Controller/ServicesprojectsController.php
gaetanHerve: ah yeah, thx

Modules/services/Controller/ServicesprojectsController.php
gaetanHerve: if I don't encode (that's what I tried first)  and get it via {{ userIsmanager }} in vuejs instance, it is equal to 1 if true and equal to "" (leads to a bug) if false.
So, I had to json_encode it to keep its type as boolean.

gaetanHerve: if I don't encode (that's what I tried first)  and get it via {{ userIsmanager }} in vuejs instance, it is equal to 1 if true and equal to "" (leads to a bug) if false.
So, I had to json_encode it to keep its type as boolean.

osallou (approved):

Issue #212, opened 2021-07-08
[Booking / Invoice] Set different night & we schedules depending on which resource is booked


gaetanHerve:
Use case raised by H2P2 :
Depending on which resource is booked, constraints on schedules are not the same.
An enhancement would be to integrate the possibility to set a specific schedule for a specific resource.

osallou:
this is an other issue (enhancement) to use a global calendar and optional per resource/resource category calendar.

osallou:
* current calendar of a space remains default
* manager can create calendar objects (identical fields to current default one)
* manager can assign resources to a calendar
* calendar has an option to be independent or shared between resources, each assigned resource bookings are independent else on booking we check they do not overlap between resources using this calendar
  * use case: 2 resources cannot be booked at the same time (in the same room for example)
* when a resource is assigned to a calendar, if was already assigned to an other one, assign to new one
* calendar view should show assigned resources
* calendar view should show "unassigned" resources (using default)


* on invoice, if resource is linked to a calendar, use this calendar for weekend/night/... calculation
* on booking if resource is linked to a calendar:
  * show and shared between resources, show all bookings for this calendar (other resources than specified one should just show a block without details), else as usual
  * show open/close for this calendar

Multiple calendar system should be flagged so that could be assigned per space plan (if plan do not have flag, just show/use default calendar)

Future enhancement idea: add per calendar closing days/hours (start/end timestamp) that would appear in calendar view as closed and checked for conflicts on booking (rather than adding fake bookings)

PR #674, opened 2022-07-01
Feature multicalendar

osallou:
Add possibility to create multiple bkscheduling (calendars)

At space create, create a default schedule
Then can select a calendar per area and/or per resource (with different times, days, ...)


osallou:
@gaetanHerve review for 2.8

gaetanHerve:
Seems to work fine.
Some details: 
- why do we have 2 default calendars ?
- maybe we could precise the name of the calendar used for a resource when it is set to "area" to have something like "area (H2P2)" based on example below

feature_multicalendar1

- I found a display alignement bug at MRic photonics':

feature_multicalendar2

osallou:
displaying name instead of area is a bit more complex, need to be updated when chaning area calendar, and is not clear if you specified this calendar for this resource, or just using the default area one

should not have 2 default, upgrade issue ? we need to create a calendar for existing areas, and is created with area name. Maybe area is named default ?

gaetanHerve:
Yes, I see it is more complex. But it is not so clear at first sight. Maybe a "parent area" instead of "area" would do the job.
About the 2 default calendars, it does that on several spaces on my side.

gaetanHerve:
and it doesn't handle well areas without resources (cf CReAAH, space 26)

feature_multicalendar3

osallou:
l'upgrade prend les calendars existant et ajoute le nom de l'area
ensuite, pour chaque espace, il créé un calendar appelé "default"

le doublon est un problème de filttrage..... je corrige

gaetanHerve:
I have some bugs testing reservations. Not sure yet it comes from this PR

osallou:
My last commits did fix the 0 calendars (deleted areas but not scheduling)

osallou:
Bookings are impacted as they have different schedules options and hours

osallou:
For the last area, will check

osallou:
"Ecochim only"  does not appear in calendar list because there is no defined calendar for this area (using default bk_scheduling object)
You can see at the bottom that area "echochim only" appears with calendar "default" as expected.

gaetanHerve:
There is still a buggy case:
When "share between resources" is active on a calendar, the display of a resource using this calendar from another resource view will display warnings (depending on what si set display to be displayed on its label I think)  and its edition won't be accessible on click (probably a resource id issue)

osallou:
pb shared calendar fixed/managed in an other issue.

gaetanHerve:
closes #212

PR #697, opened 2022-08-24
avoid displaying news popup if no news

gaetanHerve:
This PR avoid the news popup window to display if there are no news to show.


osallou:
Si testé c est ok pour moi, tu peux merger

PR #700, opened 2022-08-26
Confirm email at user creation

### LINKED ISSUE #575: [account creation] add email confirmation input in account creation forms
Some users make typos when entering their email address at account creation.
Maybe we should add an email confirmation input in account creation forms to reduce the probability of this to happen.

gaetanHerve:
closes #575


osallou:
Which case? When manager creates a user?
Cause self registratiin checks the email

osallou:
No code to manage confirm_email, where do you check it?

gaetanHerve:
It's used when a user creates their account.
The email confirmation is (indirectly) called in CoreaccountController::indexAction() line 163 with a second email field named "confirm_email" (=> framework use the name for id and name html attributes)

osallou:
Ah..  , it is not really an email confirmation, it is just adding a second email field to compare values.

It is not check the email itself like in registration.

gaetanHerve:
It is to avoid users to make an error at account creation. If they do, they'll just wait for an email that will never come...
Anyway, there is still the real "email confirmation" by clicking on the like pfm sent them. I should probably have named this PR otherwise

osallou:
if tested.... ok for me

PR #701, opened 2022-08-29
allow to delete booking authorisations, closes #647

### LINKED ISSUE #647: [booking authorizations] Add possibility to delete an authorization
Current behaviour:
In order to keep an authorizations history, it is not possible to delete an authorization (for a user to book resources within a specific category), only to set it to "inactive". Inactive authorizations will still appear in authorizations list.

Desired behaviour:
Some platforms don't need authorizations history and/or some authorization can be added by error.
So, we should make possible to delete authorizations.

gaetanHerve:


Modules/booking/Controller/BookingauthorisationsController.php
```diff
@@ -327,4 +331,22 @@ public function editAction($id_space, $id) {
...
+     * Remove a provider
+     */
+    public function deleteAction($id_space, $id) {
+        // security
+        $this->checkSpaceAdmin($id_space, $_SESSION["id_user"]);
+
+        $modelAuth = new BkAuthorization();
+
+        //get auth user
+        $authUser = $modelAuth->get($id_space, $id)["user_id"];
```
osallou, line 344: What if not found?  Access to user id will fail

Modules/booking/BookingRouting.php
```diff
@@ -15,6 +15,7 @@ public function routes($router) {
         $router->map('PROPFIND', '/caldav/[i:id_space]/[i:id_cal]', 'booking/bookingcaldav/propfind', 'booking_caldav_propfind_notrailing'); // btsync sometimes force removal of trailing slash
         $router->map('REPORT', '/caldav/[i:id_space]/1/', 'booking/bookingcaldav/report', 'booking_caldav_report');
         $router->map('REPORT', '/caldav/[i:id_space]/0/', 'booking/bookingcaldav/report', 'booking_caldav_report_default');
+        $router->map('GET|POST', '/bookingauthorisationsdelete/[i:id_space]/[i:id]', 'booking/bookingauthorisations/delete', 'booking_bookingauthorisations_delete');
```
osallou, line 18: Get or post? Only need one according to your action

gaetanHerve:
ok, I do merge

PR #705, opened 2022-08-31
Make client institutionform faculative

gaetanHerve:


osallou:
Looks ok

osallou:
Most may have same id because inserted in same order

gaetanHerve:
arf, looks like I have to modify the tests too

PR #710, opened 2022-09-14
allow to use email for login

gaetanHerve:


Modules/core/Controller/CoreconnectionController.php
```diff
@@ -143,7 +149,7 @@ private function loginError($redirection, $connection_error = "") {
                 $msg = CoreTranslator::InvalidPassword($lang);
                 break;
             case "invalid_login":
-                $msg = CoreTranslator::InvalidLogin($lang);
+                $msg = $loginType === "login" ? CoreTranslator::InvalidLogin($lang) : CoreTranslator::DuplicatedEmail($lang);
```
osallou, line 152: why duplicated email error? may just not be found

Modules/core/Model/CoreUser.php
```diff
@@ -479,45 +483,49 @@ public function isLocalUser($login) {
...
-     *      "inactive" login and password match a database entry where is_active == 0
-     *      "invalid_login" if login doesn't exist
-     *      "invalid_password" if login exists and password does not match
+     * @param string $login login
+     * @param string $pwd password
+     * @return true if connexion allowed
+     * throw PfmAuthException if unauthorised
      */
     public function connect($login, $pwd) {
         $sql = "SELECT * FROM core_users WHERE login=?";
```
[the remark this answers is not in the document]
gaetanHerve: We would still need to get the login (if user used its email) in CoreconnectionController::UserConnexion() because it is needed in LDAP part...

Modules/core/Model/CoreUser.php
```diff
@@ -475,16 +475,20 @@ public function getEmpty() {
...
-    public function isLocalUser($login) {
-        $sql = "select id from core_users where login=? AND source=?";
+    public function isLocalUser($loginOrEmail) {
+        $sql = "select * from core_users where (login=? OR email=?) AND source=?";
         $user = $this->runRequest($sql, array(
-            $login, "local"
+            $loginOrEmail, $loginOrEmail, "local"
         ));
-        return $user->rowCount() == 1;
+        if ($user->rowCount() == 1) {
```
osallou, line 487: si > 1 faudrait peut etre logger qd mm pour voir le pb

PR #716, opened 2022-09-23
make confirmation email optional at booking, closes #627

### LINKED ISSUE #627: [Booking][emails] Filter email send by resource
Would be useful to be able to set some resources to send emails (at reservation, de-reservation...) and some other resources to not send emails

gaetanHerve:


gaetanHerve:
yep, works like a charm

PR #717, opened 2022-09-23
Display client in reservations info

gaetanHerve:


Modules/booking/View/Booking/simplecaldisplay.php
```diff
@@ -122,15 +123,15 @@
 					$last_end_time = mktime($day_begin,0,0,$temp[1], $temp[2], $temp[0]);
 					foreach($calData[$calDay][$resId] as $hcalEntry) { ?>
 						<?php
-						
+						$hcalEntry['client_name'] = ($hcalEntry['responsible_id'] > 0) ? $modelClient->getName($id_space, $hcalEntry['responsible_id']) : "";
```
osallou, line 126: You make a request to get client name in a loop, for each entry. Could increase a lot number of requests.

I think client name could be fetched directly in cal entries via a join in request

Modules/booking/Model/BkCalendarEntry.php
```diff
@@ -546,6 +555,9 @@ public function getEntriesForPeriodeAndResources($id_space, $dateBegin, $dateEnd
             if(!$data[$i]["color_text"]) {
                 $data[$i]["color_text"] = "#000000";
             }
+            if(!$data[$i]["client_name"]) {
+                $data[$i]["client_name"] = "no client defined";
```
osallou, line 559: as it is a model fonction, if no client name defined, should not be replaced by a string here but remain as null (or empty).
it is in view that empty/null name should be replaced by a default string

gaetanHerve: Fixed. I found a bug. Not sure it is linked to this PR. I'm checking

Modules/booking/View/Booking/simplecaldisplay.php
```diff
@@ -123,6 +123,9 @@
 					$last_end_time = mktime($day_begin,0,0,$temp[1], $temp[2], $temp[0]);
 					foreach($calData[$calDay][$resId] as $hcalEntry) { ?>
 						<?php
+						if (!$hcalEntry['client_name']) {
+							!$hcalEntry['client_name'] = ClientsTranslator::NoCLientDefined($lang);
```
osallou, line 127: there is a typo here, ! in front of var name

Modules/booking/Model/BkCalendarEntry.php
```diff
@@ -556,7 +556,7 @@ public function getEntriesForPeriodeAndResources($id_space, $dateBegin, $dateEnd
                 $data[$i]["color_text"] = "#000000";
             }
             if(!$data[$i]["client_name"]) {
-                $data[$i]["client_name"] = "no client defined";
+                $data[$i]["client_name"] = null;
```
osallou, line 559: if not set, no need to force to null (is already empty or null)

osallou:
c'est bon de ton côté maintenant ?

un petit refactorting Modules/booking/View/Bookingconfig/summaryForm.php (éviter les répétitions des select/options) mais peut etre fait ds un second temps

gaetanHerve:
oui, tant qu'à faire... Je me mets dessus

osallou (commented):
tant qu'à faire, c'est mineur, masi mettre option et non OPTION (pas de raison de mettre en majuscules), après si testé cela me semble ok

osallou:
@gaetanHerve reste pb de merge

PR #724, opened 2022-10-05
factorize auth check in bookingdefaultController

gaetanHerve:
Forbid reservation deletion in BookingdefaultController::editreservationqueryAction if user not authorised. Was not the case before if user had access to reservation edition


Modules/booking/Controller/BookingdefaultController.php
```diff
@@ -1060,8 +1040,67 @@ public function deleteperiodAction($id_space, $id_period){
...
+            $bkAccess, $id_space,
+            $_SESSION['id_user'],
+            $curentDateUnix
+        );
+
+        $redir = $this->request->getParameterNoException('from');
+
+        $backTo = [];
+        $redirPage = '';
+        if ($redir) {
```
osallou, line 1071: should be in a try/catch in case from is not formatted as expected or do not have the 5 elts.... in case of catch, just ignore the from

Modules/booking/Controller/BookingdefaultController.php
```diff
@@ -1060,8 +1040,55 @@ public function deleteperiodAction($id_space, $id_period){
...
+        $modelBkAccess = new BkAccess();
+        $entryInfo = $modelCalEntry->getEntry($id_space, $id_calEntry);
+        $resource = $modelResource->get($id_space, $entryInfo['resource_id']);
+        $bkAccess = $modelBkAccess->getAccessId($id_space, $resource['id']);
+
+        $canValidateBooking = $this->hasAuthorization(
+            $resource['id_category'],
+            $bkAccess, $id_space,
+            $_SESSION['id_user'],
+            $entryInfo['start_time']
```
gaetanHerve, line 1056: Use new date in case of form validation / old date in case of deletion

PR #726, opened 2022-10-10
[core] add user settings to specify default number of elelments to display in arrays

osallou:
in user settings, allow user to specify the number of elements to diusplay in array by default

(all datatables elements)


gaetanHerve (approved):
Seems OK

PR #731, opened 2022-10-12
Fix block resources mandatory fields

gaetanHerve:


osallou:
did not check visual layout but looks fine for me

Modules/booking/Controller/BookingblockController.php
```diff
@@ -36,6 +36,19 @@ public function indexAction($id_space) {
         $modelColor = new BkColorCode();
         $colorCodes = $modelColor->getColorCodes($id_space);
 
+        $errorMessage = null;
+        if (empty($resources)) {
+            $errorMessage = BookingTranslator::ResourceNeeded($lang);
+        }
+        if (empty($colorCodes)) {
+            $errorMessage = $errorMessage ? $errorMessage . "</br>" : "";
+            $errorMessage .= BookingTranslator::ColorNeeded($lang);
```
osallou, line 45: if resources but no color code will do .=  on null
if resource and colorcodes, errorMessage will be null, next message will not match against "" (null is != from "")

gaetanHerve: - if resources but no color code, line 44 affects "" to $errorMessage
- ah yes, I forgot to add $erroMessage to the condition, like `if ($errorMessage && $errorMessage != "")`

Modules/booking/Model/BookingTranslator.php
```diff
@@ -1731,9 +1731,16 @@ public static function maxInvoicingUnits($lang) {
 
     public static function ColorNeeded($lang) {
         if ($lang == "fr") {
-            return "Vous devez d'abord créer au moins un code couleur dans le module  Calendrier config > Codes couleur";
+            return "Vous devez d'abord créer au moins un code couleur dans le module  Calendrier config > Codes couleur.";
         }
-        return "You need first to create at leat one color code in Bokking settings module > Color codes";
+        return "You need first to create at leat one color code in Bokking settings module > Color codes.";
```
osallou, line 1736: typo: booking

osallou (changes requested):
cf erreur typo sinon après çà, si ok pour toi ça me va

osallou (approved):

PR #747, opened 2022-11-14
[booking] add optional max booking duration Closes #741

### LINKED ISSUE #741: Booking duration
Hi,

It's not a bug, just an ideas.

On PFM, it is possible to regulate the number of reservations per day for a item.
We have resources that are used in days. 
Would it be possible to consider adding a reservation time limitation?
For example, being able to limit the reservation of a portable hard disk to 3 months?

Thanks,

Agnès

osallou:
model only for the moment not managed at booking time


osallou:
manage:

* max duration time
  * for whole booking time
  * or open hours booking time
* prevent bookings overlapping closed hours
* apply above restrictions to all or only users

PR #751, opened 2022-12-06
750 expire users on invalid email

### LINKED ISSUE #750: Expire users on invalid email
Many users, after some time, have invalid email address.

Result is there is no way to contact user, and pollute mailings

On regular basis system should send a "confirm email" check (via email...) with a confirmation delay and a link to confirm

If email is not confirmed in expected delay, user should be expired (status inactive and not used in mailings)

He should however be allowed to login to update his email and trigger a new confirmation.

osallou:
Closes #750

* Add admin button on users to check emails
* user can self check his email
* show if mail has not been checked (after upgrade) or delay has expired

New conf variables

* PFM_LANG (lang): default language (en, fr)
* PFM_EMAIL_EXPIRE_DAYS (email_expire_days): number of days before re-checking emails
* PFM_EMAIL_EXPIRE_DELAY (email_expire_delay): number of days for mail check token validity


osallou:
if expired, nothing done for the moment

osallou:
if email is unchecked/expired we should (question ?)

* on login redirect to user settings and ask user to confirm his email (check email expiration date)
* prevent emails to this address (but mail may be ok, and managers would fail to send emails....)
* prevent emails to this address for mailings but still allow for direct emails

